PR #19, opened 2019-08-27
GraphQL API docs

MonikaMrozek:
This PR adds newly created GraphQL API docs.


docs/advanced/graphql.md
```diff
@@ -3,7 +3,7 @@ id: graphql
 title: GraphQL API (Beta)
 ---
 
-> **Note** 
+> **Note**
```
NyanKiyoshi:
```suggestion
> **Note**
```

Might want a pre-commit there? I believe it works on npm. Otherwise we will end up with more of these and some editors will trail them automatically, creating _unnecessary_ diffs.

```yaml
-   repo: https://github.com/pre-commit/pre-commit-hooks
    rev: v2.2.3
    hooks:
    -   id: trailing-whitespace
    -   id: end-of-file-fixer
    -   id: flake8
```

docs/api/authenticate.md
```diff
@@ -3,4 +3,134 @@ id: authenticate
...
+If you are entering the API for the first time as a not logged in user, you will have access to the objects available only publicly. For example, by default, you can query public data, such as published products or pages.
+
+To fetch protected data, like orders or users, you need to authenticate your access.
+Saleor API uses a [JWT token](https://jwt.io/) authentication mechanism.
+Once you create a token, you need to include it as a header with every GraphQL request.
+
+The authorization header has the following format:
+
+```
+Authorization: JWT token
```
NyanKiyoshi:
```suggestion
Authorization: JWT your-token
```

docs/api/authenticate.md
```diff
@@ -3,4 +3,134 @@ id: authenticate
...
+```
+
```
NyanKiyoshi:
```suggestion

> **Note**
> You need to leave the `JWT` keyword before the token in the header value. Otherwise it will not work.

```

docs/api/authenticate.md
```diff
@@ -3,4 +3,134 @@ id: authenticate
...
+## Creating a JWT token
+
+To create a new JWT token for a user, use the `tokenCreate` mutation and provide user's credentials as the mutation input (`email` and `password`):
```
NyanKiyoshi:
```suggestion
To create a new JWT token for a user, use the `tokenCreate` mutation and provide the user's credentials as the mutation input (`email` and `password`):
```

docs/api/authenticate.md
```diff
@@ -3,4 +3,134 @@ id: authenticate
...
+  -H "Authorization: JWT your-token" \
+  -H "Content-Type: application/json" \
+  -X POST \
+  -d '{"query": "{ me { email } }"}' \
+  http://localhost:8000/graphql/
+```
+
+## Token verification and refreshing
+
+Verification and refreshing the token is straightforward. To verify the token use the following query:
```
NyanKiyoshi: Should we really tell the user it is "straightforward"?

docs/api/endpoint.md
```diff
@@ -5,7 +5,18 @@ title: The GraphQL Endpoint
...
+
+To view and try our API out, visit the [Playground](https://demo.getsaleor.com/graphql/).
+
+
+
+
+
+
+
+
```
NyanKiyoshi: We need to remove all those blank lines and only keep one (again, the `end-of-file-fixer` hook would fix those).

docs/api/error-handling.md
```diff
@@ -3,4 +3,121 @@ id: error-handling
 title: Error Handling
 ---
 
-This topic will describe the process of error handling.
\ No newline at end of file
+There are several error types in GraphQL API and you may come across several different error types, depending on the operations you are trying to perform.
```
NyanKiyoshi:
```suggestion
There are several error types in the GraphQL API and you may come across several different of them, depending on the operations you are trying to perform.
```

docs/api/error-handling.md
```diff
@@ -3,4 +3,121 @@ id: error-handling
...
+```graphql
+{
+  me {
+    email
+    fullName
+  }
+}
+```
+
+Sending this query to the server would result in the syntax error:
```
NyanKiyoshi:
```suggestion
Sending this query to the server would result in the following syntax error:
```

docs/api/error-handling.md
```diff
@@ -3,4 +3,121 @@ id: error-handling
...
+        ]
+      }
+    ]
+  }
+}
+```
+
+## Validation error
+
+This type of error occurs when user passed bad data as the mutation input. For example, while you are trying to create a new user, you are providing their email address and it happens to be already used in another user's account. It is therefore not unique and as a result you will get a validation error.
```
NyanKiyoshi:
```suggestion
This type of error occurs when user passed invalid data as the mutation input. For example, while you are trying to create a new user, you are providing their email address and it happens to be already used in another user's account. It is therefore not unique and as a result you will get a validation error.
```

docs/api/error-handling.md
```diff
@@ -3,4 +3,121 @@ id: error-handling
...
+
+Validation errors are part of the schema which means that we need to explicitly include them in the query to get them. For example in all mutations they can be obtained through the `errors` field.
```
NyanKiyoshi:
```suggestion
Validation errors are part of the schema meaning we need to explicitly include them in the query to get them. For example in all mutations they can be obtained through the `errors` field.
```

docs/api/error-handling.md
```diff
@@ -3,4 +3,121 @@ id: error-handling
...
+    edges {
+      node {
+        id
+      }
+    }
+  }
+}
+```
+
+Executing it without proper permission would result in the error:
```
NyanKiyoshi:
```suggestion
Executing it without proper permission would result in the following error:
```

docs/api/example-query.md
```diff
@@ -3,9 +3,10 @@ id: example-query
 title: Example Query - Making your First Call
 ---
 
-## Example query
+Requests must be sent using HTTP `POST` method and `application/json` content type.
```
NyanKiyoshi:
```suggestion
Requests must be sent using the HTTP `POST` method and the `application/json` content type.
```

docs/api/example-query.md
```diff
@@ -3,9 +3,10 @@ id: example-query
...
 
-Querying data in GraphQL is easy with tool GraphiQL. The tool can be used from a web browser.
+
+There are two ways of querying API. You can either use the curl library, which is a low level developer's tool, or use Playground, an interactive editor from GraphQL API.
```
NyanKiyoshi:
```suggestion
There are two ways of querying the API. You can either use the cURL library, which is a low level developer's tool, or use Playground, an interactive editor from GraphQL API.
```

**or**

```suggestion
There are two ways of querying Saleor API. You can either use the cURL library, which is a low level developer's tool, or use Playground, an interactive editor from GraphQL API.
```

docs/api/playground.md
```diff
@@ -3,4 +3,15 @@ id: playground
 title: How to Access the Playground
 ---
 
-This topic will describe the process of accessing and using playground.
\ No newline at end of file
+GraphQL Playground is external interactive editor for your GraphQL queries. It is based on the GraphiQL and accessible through the web browser.
```
NyanKiyoshi:
```suggestion
GraphQL Playground is an external interactive editor for your GraphQL queries. It is based on the GraphiQL and accessible through the web browser. 
```

docs/api/playground.md
```diff
@@ -3,4 +3,15 @@ id: playground
 title: How to Access the Playground
 ---
 
-This topic will describe the process of accessing and using playground.
\ No newline at end of file
+GraphQL Playground is external interactive editor for your GraphQL queries. It is based on the GraphiQL and accessible through the web browser. 
+
+With the `DEBUG=True` setting enabled, Saleor exposes an interactive GraphQL editor under `/graphql`, which allows accessing the API from your browser. To view and try our API out, visit the [Playground](https://demo.getsaleor.com/graphql/).
```
NyanKiyoshi:
```suggestion
With the `DEBUG=True` setting enabled, Saleor exposes an interactive GraphQL editor under `/graphql`, allowing to access the API from your browser. To view and try out our API, visit the [Playground](https://demo.getsaleor.com/graphql/).
```

docs/api/playground.md
```diff
@@ -3,4 +3,15 @@ id: playground
...
+
+Using Playground is easy and intuitive with its sidebar navigation containing two tab (_Docs_ and _Schema_), listing all operations available in Saleor API. The Playground allows you to quickly familiarize yourself with API, perform example operations and send your first queries.
+
+
+
+SCREENSHOT OF PLAYGROUND SALEOR WELCOME MESSAGE
```
NyanKiyoshi:
```suggestion
<!-- FIXME: SCREENSHOT OF PLAYGROUND SALEOR WELCOME MESSAGE -->
```

docs/api/authenticate.md
maarcingebala: There are at least three examples, I think it should be clear.

docs/api/example-query.md
```diff
@@ -34,30 +35,76 @@ The server returns the following JSON:
...
+
+Let's look at an example of a mutation. Imagine that we want to update the name of a product. In Saleor API this can be done with the `productUpdate` mutation:
+
+```graphql
+mutation {
+  productUpdate(
+    id: "UHJvZHVjdDo3Mg=="
+    input: { name: "Apple Juice - Summer Edition" }
+  ) {
+    errors {
```
NyanKiyoshi: Shouldn't we keep the "convention" of putting the errors at the end of the query?

docs/api/authenticate.md
patrys: @MonikaMrozek Let's avoid qualifiers like "easy", "straightforward", "simple", etc. They don't make things any easier and lead to pointless frustration when things are not as obvious to the reader.

docs/api/example-query.md
```diff
@@ -3,20 +3,22 @@ id: example-query
 title: Example Query - Making your First Call
 ---
 
-## Example query
+There are two ways of querying API. You can either use the cURL library, which is a low level developer's tool, or use Playground, an interactive editor from GraphQL API.
```
patrys: I think this is slightly misleading, there are definitely lots of other ways to query the API. For production code, you would usually use a dedicated GraphQL client, not curl. Curl is just a way to play with the API from your command line.

docs/advanced/graphql.md
maarcingebala: It may be a good idea. Unfortunately, I won't have time now to configure it. Correct me if I'm wrong, but I thought pre-commit only works when creating commits from shell? None of our writers use shell but the UI tools for git.

NyanKiyoshi: It works but they might run into weird behaviors from the editors as they most often don't expect such responses.

Issue #18, opened 2019-08-26
Wrong routing


taoufik07:
Some of the cards links in the docs returns a 404 but when I try to access via the sidebar it works.

I've noticed that the non working links on the cards e.g. points to `docs/customization/intro/customization/docker.md`  while on the sidebar it points to `docs/customization/docker/`

NyanKiyoshi:
cc. @rafalp

PR #22, opened 2019-08-29
fixed urls

MonikaMrozek:


rafalp:
This PR fixes #18 by changing Docusaurus links in documents to absolute URL, working around the bug in link-transformation logic only supporting docs that exist in `/docs/` and ignoring historical documents from `versioned_docs`.

When we will be releasing next version of the docs, we will have to remember to run find&replace on `versioned_docs/2.8.0/` to replace `/docs/` to `/docs/2.8.0/` so they remain active, cc @maarcingebala  & @MonikaMrozek

PR #24, opened 2019-09-04
Checkout process topic

MonikaMrozek:
please, add your code snippets

### DIFF
--- a/docs/api-process/check-out.md
+++ b/docs/api-process/check-out.md
@@ -3,4 +3,76 @@ id: check-out
 title: How to Create Checkout?
 ---
 
-This topic describes the process of check out.
+Below is a description of a checkout process. We assume that at this stage you have already completed the steps included in the Getting Started section of this chapter and you are familiar with basic setup of Saleor GraphQL API.
+
+The code snippets included in this section may be run in ![Playground](api/playground.md) or your preferred HTTP client.
+
+Checkout process in Saleor may run through several process flow scenarios, depending on such factors like, for example, if the user is logged in or not at the moment of adding items to the cars. 
+
+Saleor does not create an ‘empty checkout’ when a user is not logged in but adds items to the cart. In these situations the cart is not saved anywhere on the server but on the browser level.
+
+The actual checkout process is started after the user clicks check out in their cart. If the user is already logged in - the process in commenced. If not, the user is prompted to log into their account.
+
+## Step #1 - mutation: `CHECKOUT CREATE`
+
+To create a checkout object in the database, execute the following mutation: checkout create. 
+
+This mutation takes the following input:
+
+* User’s email address
+
+* User’s shipping address
+
+* User’s billing address
+
+* List of checkout lines (each checkout line represents the variant id, in other words - the product, and the quantity of it)
+
+There are also other fields in this mutation:
+
+* Checkout id
+
+* Checkout token - this is a unique token for this checkout which allows you to identify this specific checkout object. This is a public token, it will save the user’s checkout session, if for example, they accidentally close the browser.
+
+* Total price of the items in the cart
+
+> **Note**
+>
+> An email is required to create the checkout object. 
+
+The following example shows how the mutation creates the checkout object and returns the checkout id: 
+
+<!-- Marcin to provide a code snippet from Playground -->
+
+### `available shipping methods`
+
+The `CHECKOUT CREATE` object returns several fields. The `available shipping methods` is one of these fields (this is an optional step, for items which require shipping only).
+
+Each shipping method on the list in this fields has a unique id. 
+
+If the items in the cart requires shipment, the user will select the shipping method. 
+
+To allow the user to effectively select a specific shipping method (to create a shipping in this checkout) you need to run the following operation: `checkout shipping method update`.
+
+This operation has two parameters:
+
+* Checkout id
+
+* Shipping method id
+
+This way, the specific checkout is paired up with the specific shipping method selected by the user. 
+
+<!-- Marcin to provide a code snippet from Playground -->
+
+## Step #2 - Creating payment
+
+The checkout object contains the `available payment gateways` field. It returns a list of the payment gateways which are currently configured on your Saleor server.
+
+Payment creation process consists of two operations:
+
+* Generating a token for this specific payment using the `payment client token` operation. 
+This operation requires the user to select the preferred payment gateway.
+
+* Executing mutation `CHECKOUT PAYMENT CREATE` using the token generated in the previous step. 
+
+
+
--- a/docs/api-process/check-out.md
+++ b/docs/api-process/check-out.md
@@ -31,7 +31,7 @@ This mutation requires the following input:
 
 - `billingAddress` - user's billing address
 
-- `lines` - list of checkout lines - each checkout line represents the _variant id_ (the specific product) and the quantity of it.
+- `lines` - list of checkout lines - each checkout line represents the _variant id_ (the specific product) and the quantity of it
 
 As a result, this mutation returns the following fields:
 
--- a/docs/api-process/check-out.md
+++ b/docs/api-process/check-out.md
@@ -5,17 +5,21 @@ title: How to Create Checkout?
 
 Below is a description of a checkout process. We assume that at this stage you have already completed the steps included in the Getting Started section of this chapter and you are familiar with basic setup of Saleor GraphQL API.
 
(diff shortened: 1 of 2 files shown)


docs/api-process/check-out.md
```diff
@@ -3,4 +3,108 @@ id: check-out
...
+
+Depending on selected payment gateway, you will either use the JavaScript form which can be integrated to Saleor, or the payment gateway directs you to an external payment page.
+
+JavaScript responds returning a `promise` object. The payment gateway sends an information if the payment is successful along with a tokenized credit card payment information. 
+
+This tokenized card information is then used to run the `checkout payment create` operation in order to save this payment record in API database.
+
+### `checkout payment create`
+
+The `checkout payment create` mutation requires the following three parameters:
```
NyanKiyoshi: I think it would be nice to have a note that this operation only stores the tokenized information, but never it will try to charge or authorize the payment by itself. (that's `checkoutComplete`'s job)

docs/api-process/check-out.md
```diff
@@ -3,4 +3,108 @@ id: check-out
...
+
+## Step #3 Complete checkout
+
+This operation requires only the checkout id. Its purpose is to ensure this checkout is correct and in order to do that, it verifies if:
+
+* The addresses are correct 
+
+* The products are in fact in stock (while making the purchase, another user could already buy the last available item)
+
+* The payment has been successful
```
NyanKiyoshi:
```suggestion
* The payment has been successfully charged
```

docs/api-process/check-out.md
```diff
@@ -3,4 +3,108 @@ id: check-out
...
+
+If these parameters are verified correctly, then the checkout is transformed into an order. In the same time, the customer receives an email with a confirmation of placing an order.
```
NyanKiyoshi: In or at the same time?

docs/api-process/check-out.md
```diff
@@ -3,4 +3,108 @@ id: check-out
...
+
+The below process describes the key milestones in the checkout process flow in Saleor. There are also additional steps that may occur along the way. However, the purpose of this instruction is to deliver a base reference for the user to work with.    
+
+The code snippets included in this section may be run in ![Playground](api/playground.md) or your preferred HTTP client.
+
+### Empty checkout
+
+The checkout process in Saleor may run through several scenarios depending on such factors like, for example, if the user is logged in or not, at the moment of adding items to the cart. 
+
+Saleor does not create ‘empty checkouts’ when a not-logged-in user is adding items to the cart. In such instances, the current cart is only saved on the browser level. It is not saved anywhere on the server.
```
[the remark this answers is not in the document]
maarcingebala: All of the work we're doing will be reviewed by @timcla before the release, so we don't have to bother with every little grammar issue ;)

maarcingebala:
@NyanKiyoshi thanks for the review, but I think this is the very first version of the document. We work on those technical docs in two parts - first Monika lays out the overall structure and most of the text, then someone from the technical team adds all the details such as mutation names, code snippets, etc.

NyanKiyoshi:
Okay, thanks :+1:

docs/api-process/check-out.md
```diff
@@ -3,4 +3,108 @@ id: check-out
...
+
+The actual checkout process is started after the user clicks _Checkout_ in their cart. If the user is already logged in - the process is commenced. If not, the user is prompted to log into their account first.
```
bogdal: I think, it is worth mentioning that the login process is optional. The user can continue as an anonymous user.

docs/api-process/check-out.md
```diff
@@ -1,6 +1,119 @@
 ---
 id: check-out
-title: How to Create Checkout?
+title: How to Create Checkout
 ---
+## Introduction
+
+Below is a description of a checkout process. We assume that at this stage you have already completed the steps included in the [Getting Started](api/intro.md) section of this chapter and you are familiar with basic setup of Saleor GraphQL API.
```
NyanKiyoshi:
```suggestion
Below is a description of a checkout process. We assume that at this stage you have already completed the steps included in the [Getting Started](api/intro) section of this chapter and you are familiar with basic setup of Saleor GraphQL API.
```



I believe that's how we are supposed to link them? Can you check?

docs/api-process/check-out.md
```diff
@@ -1,6 +1,349 @@
...
+>
+> - If no authentication token is provided, the checkout is created for an anonymous user, and an email address is used to identify such `Checkout` object linking it with the anonymous user. In this case, an email is required to create checkout.
+
+To create a `Checkout` object, use the `checkoutCreate` mutation.
+
+This mutation requires the following input:
+
+- `email` - user's email address
+
+- `shippingAddress` - user's shipping address
```
korycins: Shouldn't we add the name of the structure? Could be useful for developers to easier find it in schema. Same question for all complex structure eg `lines` .

docs/api-process/check-out.md
```diff
@@ -1,6 +1,349 @@
...
+```
+
+As a result we're getting the payment object:
+
+```json
+{
+  "data": {
+    "checkoutPaymentCreate": {
+      "payment": {
+        "id": "",
```
korycins: Should `id` be empty?

docs/api-process/check-out.md
maarcingebala: You mean the type of the input e.g. `AddressInput`?

docs/api-process/check-out.md
maarcingebala: No, good catch.

docs/api-process/check-out.md
korycins: Yes, it would be helpful for me to find it in the schema.  Adding a type of any complex structure would be a good idea.

maarcingebala: We could do that but we don't have API reference. When we have it, we will create links from guides to those specifications.

korycins: Sounds good.

docs/api-process/check-out.md
```diff
@@ -1,6 +1,349 @@
...
+      },
+      "errors": []
+    }
+  }
+}
+```
+
+## Step #2 (optional) - `checkoutShippingMethodUpdate`
+
+This step is only used if purchased items require shipping (if they are physical products). The user must select a specific shipping method to create shipping for this checkout.
```
krzysztofwolski: Do we have a way to check if checkout _needs_ shipping?

NyanKiyoshi: Yup! `checkout.isShippingRequired`

docs/api-process/check-out.md
```diff
@@ -1,6 +1,349 @@
...
+
+This mutation requires the following input:
+
+- `email` - user's email address
+
+- `shippingAddress` - user's shipping address
+
+- `billingAddress` - user's billing address
+
+- `lines` - list of checkout lines - each checkout line represents the _variant id_ (the specific product) and the quantity of it.
```
bogdal: It would be nice to use one format for all items in the list. I suggest getting rid of the ending dot and using a semicolon to separate sentences.

```suggestion
- `lines` - list of checkout lines - each checkout line represents the _variant id_ (the specific product) and the quantity of it
```

docs/api-process/check-out.md
patrys: But you have to create the checkout object first, we'll eventually get to documenting how you can add addresses later.

krzysztofwolski: Would be awesome if we mention it here

docs/api-process/check-out.md
```diff
@@ -1,6 +1,349 @@
...
+## Step #1 - `checkoutCreate`
+
+> **Note**
+>
+> A `Checkout` object can be created for logged in users and for anonymous (guest) users.
+>
+> - If you use the `checkoutCreate` mutation including the authentication token, this checkout is assigned to the user who is authenticated by this token. For more information on how to authenticate with our API, see the [Authentication](api/authenticate.md) topic.
+>
+> - If no authentication token is provided, the checkout is created for an anonymous user, and an email address is used to identify such `Checkout` object linking it with the anonymous user. In this case, an email is required to create checkout.
+
```
fowczarek: I'm not sure if we should describe the situation that creating checkout as an anonymous user with an email of an existing customer.

fowczarek:
It would be nice to use one schema for `Step` titles. In step 1, 2, 4 we heave mutation name as title and description of one mutation, but in step 3 we have mutation name as title and description of two requests(paymentClientToken, checkoutPaymentCreate).

PR #36, opened 2019-09-13
metadata topic added

MonikaMrozek:
Added topic describing extending Saleor using metadata.


docs/api/metadata.md
```diff
@@ -0,0 +1,219 @@
...
+title: Extending Models Using Metadata
+---
+
+Saleor gives you the possibility to customize your shop without the need of any changes in the code. Once you get your own Saleor instance, you can extend most of Saleor's database models using metadata. 
+
+For example, you can add specific fields to metadata of objects to store extra information on that object. Using metadata you can also customize the way your storefront appears to your customers. 
+
+The following models contain metafields which data can be modified:
+
+* `collection`
```
maarcingebala: I would use `PascalCase` for model names, that is the convention in Django and in GraphQL when referring to objects.

docs/api/metadata.md
```diff
@@ -0,0 +1,219 @@
+---
+id: metadata
+title: Extending Models Using Metadata
+---
+
+Saleor gives you the possibility to customize your shop without the need of any changes in the code. Once you get your own Saleor instance, you can extend most of Saleor's database models using metadata. 
+
+For example, you can add specific fields to metadata of objects to store extra information on that object. Using metadata you can also customize the way your storefront appears to your customers.
```
maarcingebala: Shouldn't it be:
```suggestion
For example, you can add specific fields to metadata of objects to store extra information on that object. By using metadata you can also customize the way your storefront appears to your customers. 
```

docs/api/metadata.md
```diff
@@ -0,0 +1,219 @@
...
+                  {
+                    "name": "vatlayer",
+                    "metadata": [
+                      {
+                        "key": "code",
+                        "value": "standard"
+                      },
+                      {
+                        "key": "description",
+                        "value": ""
```
maarcingebala: @korycins Can we add some example description here?

docs/api/metadata.md
```diff
@@ -0,0 +1,219 @@
...
+
+```
+mutation {
+  productUpdateMetadata(id:"UHJvZHVjdDo3Mg==", input:{namespace:"my_namespace1", clientName: "custom_name", key:"special_product", value:"true"}){
+    errors{
+      field
+      message
+    }
+    product{
+		meta{
```
maarcingebala: @korycins Please format the example queries properly. You can use "Prettify" option in the Playground. Not only this one but all of them, some of them have pretty long attributes lists.

docs/api/metadata.md
```diff
@@ -0,0 +1,219 @@
...
+}
+```
+
+## Private and public metafields
+
+Models can have two meta fields:
+
+* `meta` - This field is public and visible for the non-authenticated user
+
+* `privateMeta` This  is protected and visible only for staff user with expected permissions
```
maarcingebala:
```suggestion
* `privateMeta` This is protected and visible only for staff user with expected permissions
```

docs/api/metadata.md
MonikaMrozek: My bad! I've asked Maciek to paste the code snippets into the task in Jira, so I can place them in the docs. @korycins looks like you will have to redo them anyway :( sorry for the doubkle work!

PR #39, opened 2019-09-17
Pagination

MonikaMrozek:


docs/api/pagination.md
```diff
@@ -0,0 +1,145 @@
...
+    }
+  }
+}
+```
+
+* `before` - Followed by the value of the `cursor`. This indicates the direction in which you want to start displaying items, in this case - from the end of the list. 
+
+```
+{
+  products(first: 10, before: "YXJyYXljb25uZWN0aW9uOjQ==") {
```
maarcingebala: `before` is used with `last`:

```suggestion
  products(last: 10, before: "YXJyYXljb25uZWN0aW9uOjQ==") {
```

PR #46, opened 2019-09-25
Feature new docs design

benekex2:
screencapture-localhost-3001-docs-2019-09-25-20_28_16


maarcingebala:
- [ ] Top right links are wrong - the first one should link to the dashboard guide and be renamed according to designs; the last one should link to the developer's guide. Also on design there is a bit different font.

image

- [ ] When going to "GraphQL API" docs in the top right nav, there is no underline as it is for other sections.

- [ ] I think the spacing should be bigger between the footer and the last element:

image

- [ ] On some widths layout breaks:
image

- [ ] Github icon is missing eyes 👀

- [ ] Missing links in footer icons / links

PR #55, opened 2019-10-08
Add copy button on code block

lex111:
AwesomeScreenshot-2019-10-8-1570527481822

### DIFF
--- a/website/siteConfig.js
+++ b/website/siteConfig.js
@@ -101,8 +101,13 @@ const siteConfig = {
   docsSideNavCollapsible: false,
 
   // Add custom scripts here that would be placed in <script> tags.
-  scripts: ["https://buttons.github.io/buttons.js", "/js/script.js"],
-
+  scripts: [
+    "https://buttons.github.io/buttons.js",
+    "https://cdnjs.cloudflare.com/ajax/libs/clipboard.js/2.0.0/clipboard.min.js",
+    "/js/code-block-copy-button.js",
+    "/js/script.js"
+  ],
+  
   // On page navigation for the current documentation page.
   onPageNav: "separate",
   // No .html extensions for paths.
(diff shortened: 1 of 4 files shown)


maarcingebala:
This is so cool!

maarcingebala:
Next level - running GraphQL queries inside the docs. I haven't looked into this yet and I don't know if this is possible in Docusaurus. But would be cool :)

lex111:
So it does not depend on Docusaurus, should the server execute the GraphQL query? I can’t even imagine how it’s possible, I saw how in the docs putting a link to the sandbox where the query (or sample code) is execute.

maarcingebala:
About the "copy" button - there seems to be an "official" solution that was created by the maintainer of Docuzaurus as a gist. Can we use his solution? How is yours different from it?

lex111:
As I understand it, do you mean this [gist](https://gist.github.com/yangshun/55db997ed0f8f4e6527571fc3bee4675)?
I actually did the same, but improved it. When copying the bash command, the `$` sign is removed from the clipboard, in my opinion it is very convenient.
You can even simply compare the code, it is almost the same.

Offtop: I noticed that there is no `og-image` on the docsite, I would like to add it, but for this I need the image itself (1200x630), can you send it to me?

maarcingebala:
@lex111 Please open an issue about this image, I don't have it currently but we'll get back to it when we have time :)

lex111:
@maarcingebala opened https://github.com/mirumee/saleor-docs/issues/58

NyanKiyoshi:
@lex111 maybe we could hide the outline of the button? Happens after we click on it
Screenshot 2019-10-08 at 23 19 57

lex111:
@maarcingebala sure! But there shouldn’t be this outline, because I added some CSS code.. apparently I need to add similar one for `click`.

https://github.com/mirumee/saleor-docs/blob/1a8c7934efbf932f5c0473b2b0be8f2298398d30/website/static/css/code-block-copy-button.css#L20-L21

PR #64, opened 2019-10-10
Add unified button style

lex111:
Currently styles for buttons (edit button) with a green border  due the default theme style for Docusaurus. This PR overrides the default styles for the button to look like other buttons on docsite.

Before:

image

After:

image


maarcingebala:
Seems legit, thanks for taking care!

PR #92, opened 2020-01-02
Updated Linux Installation documentation

aselam:
Added a few warnings related to errors stumbled upon during Saleor 2.9 installation:
* Python and Pip 3.8
* Front-end assets can't be built as they are now in separate repositories.

Should there be documentation regarding Dashboard and Storefront installation here?

### DIFF
--- a/docs/getting-started/installation-linux.md
+++ b/docs/getting-started/installation-linux.md
@@ -87,6 +87,11 @@ We strongly recommend [creating a virtual environment](https://docs.python.org/3
 $ pip install -r requirements.txt
 ```
 
+> **Warning**
+>
+> If you are getting errors related to missing hashes, please update your Python and Pip to a more recent version (3.8+ recommended).
+>
+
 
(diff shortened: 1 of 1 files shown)


docs/getting-started/installation-linux.md
```diff
@@ -150,6 +157,11 @@ $ npm install
 $ npm run build-assets
 ```
 
+> **Warning**
+>
+> Since version 2.9, it may be not working anymore because [Dashboard](https://github.com/mirumee/saleor-dashboard) and [Storefront](https://github.com/mirumee/saleor-storefront) are available through separate repositories :
```
NyanKiyoshi: Would maybe sound better like so?
```suggestion
> Since version 2.9, it may not be working anymore as [Dashboard](https://github.com/mirumee/saleor-dashboard) and [Storefront](https://github.com/mirumee/saleor-storefront) are available through separate repositories : 
```

PR #102, opened 2020-02-17
Update email compile related docs

tomaszszymanski129:
As we discussed we should provide pre-compiled email templates instead of forcing user to compile them on start. Should be merged along with backend changes on https://github.com/mirumee/saleor/pull/5260.

### DIFF
--- a/docs/customization/emails.md
+++ b/docs/customization/emails.md
@@ -21,7 +21,18 @@ Emails defined in `*.email` files include their HTML versions by referencing the
 
 Source emails use [MJML](https://mjml.io/) and must be compiled to HTML before use.
 
-To compile emails, run:
+### Before you start
+
+To compile the emails you need to have node.js installed.
+Version 10 or later is required. See the [installation instructions](https://nodejs.org/en/download/package-manager/).
+
+You also need to install frontend dependencies first, run:
+
+```console
+$ npm install
+```
+
+To compile the emails run:
 
 ```console
 $ npm run build-emails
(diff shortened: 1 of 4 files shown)


docs/customization/emails.md
```diff
@@ -21,7 +21,18 @@ Emails defined in `*.email` files include their HTML versions by referencing the
 
 Source emails use [MJML](https://mjml.io/) and must be compiled to HTML before use.
 
-To compile emails, run:
+### Before you start
+
+To compile the emails you need to have node.js installed.
```
maarcingebala:
```suggestion
To compile the emails you need to have Node.js installed.
```

docs/customization/emails.md
```diff
@@ -21,7 +21,18 @@ Emails defined in `*.email` files include their HTML versions by referencing the
...
+Version 10 or later is required. See the [installation instructions](https://nodejs.org/en/download/package-manager/).
+
+You also need to install frontend dependencies first, run:
```
maarcingebala: I suggest a little rephrase to make it more natural

```suggestion
You also need to install frontend dependencies first. To do that run:
```

docs/customization/emails.md
```diff
@@ -23,10 +23,10 @@ Source emails use [MJML](https://mjml.io/) and must be compiled to HTML before u
 
 ### Before you start
 
-To compile the emails you need to have node.js installed.
+To compile the emails you need to have Node.js installed.
 Version 10 or later is required. See the [installation instructions](https://nodejs.org/en/download/package-manager/).
```
NyanKiyoshi: Node JS 8 or greater actually (please check, but AFAIK that's the MJML requirement) which would be easier for non-rolling release distribution users.

maarcingebala: Node.js 8 is no longer supported: https://github.com/nodejs/Release#release-schedule. I guess it's fine as it is.

PR #103, opened 2020-02-20
Modify docker installation notes for saleor platform

tomaszszymanski129:
Docker installation notes, changes according to
https://github.com/mirumee/saleor-platform/pull/1
https://github.com/mirumee/saleor/pull/5278

### DIFF
--- a/docs/getting-started/docker.md
+++ b/docs/getting-started/docker.md
@@ -26,6 +26,6 @@ In order for Docker to use your assets from the host, you need to remove `/app/s
 Additionally, if you wish to have the compiled email templates mounted, also remove the `/app/templates/templated_email/compiled` volume from the web and celery services.
 
 
-## Use Docker to run saleor, storefront and dashboard at once
+## Use Docker to run Saleor, storefront and dashboard at once
 
  Navigate to [saleor-platform](https://github.com/mirumee/saleor-platform/) repository, it contains platform dependent instructions to get it up and running in no time.
--- a/docs/getting-started/docker.md
+++ b/docs/getting-started/docker.md
@@ -14,48 +14,18 @@ Using Docker to build software allows you to run and test code without having to
 
 You will need to install [Docker](https://docs.docker.com/install/) and [docker-compose](https://docs.docker.com/compose/install/) before following the instructions below.
 
-> **Note**
->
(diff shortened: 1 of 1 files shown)


docs/getting-started/docker.md
```diff
@@ -14,48 +14,18 @@ Using Docker to build software allows you to run and test code without having to
...
->**Note**
-> 
->The `--createsuperuser` argument creates an admin account for `admin@example.com` with the password set to `admin`.
-
-**3. Run the containers**
-
-```console
-$ docker-compose up
-```
+## Use Docker to run saleor, storefront and dashboard at once
```
maarcingebala:
```suggestion
## Use Docker to run Saleor, storefront and dashboard at once
```

docs/getting-started/docker.md
```diff
@@ -14,48 +14,18 @@ Using Docker to build software allows you to run and test code without having to
...
-
-If you do not wish to use any overrides, you can order compose to only use docker-compose.yml configuration using `-f`, for example: `docker-compose -f docker-compose.yml up`.
-
 
 ## Using local assets
 
 By default, we do not mount assets for development in the Docker. This is because the assets are built in the Docker at build-time and are not present in the cloned repository. As a result, what has been built on the Docker, would be subsequently overshadowed by empty directories from the host.
 
-However, we are aware that in some instances you may wish to mount the assets and see your changes reflected in the container. If so, modify the docker-compose.override.yml before proceeding.
+However, we are aware that in some instances you may wish to mount the assets and see your changes reflected in the container. If so, modify the docker-compose.yml before proceeding.
```
NyanKiyoshi: Wouldn't you want to make a `docker-compose.override.yml` instead? Seems like a better practice, but maybe harder for new comers to Docker

NyanKiyoshi: You could link that documentation link back here as a note: https://docs.docker.com/compose/extends/#understanding-multiple-compose-files

And add the command to run it:
```
docker-compose -f docker-compose.yaml -f docker-compose.override.yaml up
```

tomaszszymanski129: @NyanKiyoshi command suggested can be simplified to `docker-compose up` - if these files are named as you suggested, compose will take it all into consideration.

The reason behind moving only compose file here (merge Saleors compose and override files + added some extra services) and dropping the override is that we use compose for only one environment - development.

Since there is no other env for compose I don't see a need for override file, furthermore with override added it can be confusing for new Docker new comers (which might not know that override is magically taken into consideration).

PR #107, opened 2020-02-26
Add documentation for OpenTracing using Jaeger

tomaszszymanski129:
Notes about features of OpenTracing / Jaeger and what do we trace inside Saleor.

IMPORTANT: Should be merged along with https://github.com/mirumee/saleor/pull/5300


docs/guides/opentracing-jaeger.md
```diff
@@ -0,0 +1,54 @@
+---
+id: opentracing-jaeger
+title: OpenTracing with Jaeger
+---
+
+Saleor has an implementation of OpenTracing used for tracing performance of the application. Jaeger is the currently implemented provider which allows to recap whats collected through dedicated web interface.
```
NyanKiyoshi:
```suggestion
Saleor has an implementation of OpenTracing used to trace performance of the application. Jaeger is the currently implemented provider which allows to recap what is collected through a dedicated web interface.
```

docs/guides/opentracing-jaeger.md
```diff
@@ -0,0 +1,54 @@
...
+
+Jaeger web interface helps you see through particular request, check how many queries it executes underneath, what these queries are (both SQL and GraphQL) and how much time each particular part of the process takes, all on a timeline of a single request.
```
NyanKiyoshi:
```suggestion
Jaeger web interface helps you see through particular requests, such as checking how many queries it executes underneath, what these queries are (both SQL and GraphQL), how much time each particular part of the process takes, all on a timeline of a single request and much more.
```

This is not limited to only those few things, the tracing is still a work in progress, more will come and it will improve 😄 On Jaeger you can see a map/chart as well, which is a neat and useful feature, not only the segments themselves!

docs/guides/opentracing-jaeger.md
```diff
@@ -0,0 +1,54 @@
...
+    - tags
+        - component - "graphql_query"
+        - error - optional, True if error occured
+    - logs
+        - query - querystring with length limited to settings.OPENTRACING_MAX_QUERY_LENGTH_LOG
+- PostgreSQL queries using Django middleware
+    - tags
+        - component - "db"
+        - db.type - "PostgreSQL"
+        - db.statement - SQL query statement
```
NyanKiyoshi: I think it would belong to its own opentracing section than jaeger (and adding a link to it here)
And the documentation is already outdated 😉 https://github.com/mirumee/saleor/pull/5292

NyanKiyoshi: All of those segments can have `error`, `stack`, `error.kind`, `error.object` if a segment happens to fail

tomaszszymanski129: @NyanKiyoshi is this already implemented? i can see `error` implemented but not the other ones

NyanKiyoshi: They are yes, this is done underneath, whenever a span crashes, opentracing will report the error.

docs/guides/opentracing-jaeger.md
maarcingebala: Grammarly gave me this suggestion for the original text:

"Saleor has an implementation of OpenTracing used for tracing the performance of the application. Jaeger is the currently implemented provider that allows recapping what's collected through a dedicated web interface."

NyanKiyoshi: Sounds good

docs/guides/opentracing-jaeger.md
```diff
@@ -0,0 +1,25 @@
+---
```
NyanKiyoshi: I think you could put it under advanced: https://docs.saleor.io/docs/next/advanced/intro/ as it is meant for developers

PR #111, opened 2020-03-18
ExtensionsManager was moved out of core

reggaemuffin:
It seems like ExtensionsManager is not in core anymore so I wanted to update the docs to correct this :)


maarcingebala:
Good catch, thanks!

PR #132, opened 2020-05-26
Fetch manifest section

korycins:
Add fetchManifestSection

### DIFF
--- a/docs/developer/extending/apps.mdx
+++ b/docs/developer/extending/apps.mdx
@@ -46,10 +46,30 @@ Saleor expectes the manifest to use the following format:
   "supportUrl": "http://localhost:3000/support"
 }
 ```
+:::note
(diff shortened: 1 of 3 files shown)


docs/developer/extending/apps.mdx
```diff
@@ -1,8 +1,532 @@
...
+        "appName": "Order App",
+        "manifestUrl": "http://localhost:3000/manifest"
+      },
+      "appErrors": []
+    }
+  }
+}
+```
+
+The chart below explains workflow of the installation process:
```
IKarbowiak, line 290: Probably `chart` should be in plural.

patrys:
Is this API useful to anyone but the dashboard? If not, are we starting to document dashboard-specific APIs?

korycins:
> Is this API useful to anyone but the dashboard? If not, are we starting to document dashboard-specific APIs?

@patrys  You are right. I didn't think about this. So I assume that we can close it.

PR #133, opened 2020-05-28
Document plugin discovery

patrys:
This is the documentation change related to https://github.com/mirumee/saleor/pull/5686

### DIFF
--- a/docs/developer/extending/plugins.mdx
+++ b/docs/developer/extending/plugins.mdx
@@ -111,7 +111,7 @@ Saleor uses the *entry points* API of Python's `setuptools` to automatically dis
 from setuptools import setup
 
 setup(
-    ...
+    ...,
     entry_points={
         "saleor.plugins": [
             "my_plugin = my_plugin.plugin:MyPlugin"
--- a/docs/developer/extending/plugins.mdx
+++ b/docs/developer/extending/plugins.mdx
@@ -6,7 +6,7 @@ title: Plugins
 
 Plugins offer a mechanism to extend Saleor's Python core with new functionality. They are loaded by the core process and use predefined callbacks to execute custom code in response to certain events.
 
-Saleor ships with some plugins included, you can find those in the `saleor/plugins/` directory of the core.
+Saleor ships with some plugins included, you can find those in the `saleor/plugins/` and `saleor/payment/gateways/` directories of the core.
 
 ## Base plugin class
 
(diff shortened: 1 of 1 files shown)


docs/developer/extending/plugins.mdx
```diff
@@ -4,20 +4,15 @@ title: Plugins
...
-Saleor arrives with some plugins already implemented by default. These plugins are located in `saleor.plugins`.
-To provide the `PluginsManager` with a list of enabled plugins, include the Python plugin path in the `settings.PLUGINS` list.
+Plugins offer a mechanism to extend Saleor's Python core with new functionality. They are loaded by the core process and use predefined callbacks to execute custom code in response to certain events.
 
-## Plugins manager
-
-The `PluginsManager` is located in the `saleor.plugins.manager`. It is a class responsible for handling all declared plugins and serving a response from them. In case of a non-declared plugin, it serves a default response.
-
-It is possible to overwrite an `PluginsManager` class by implementing it on its own. Saleor will discover the manager class by taking the declared path from `settings.PLUGINS_MANAGER`.
+Saleor ships with some plugins included, you can find those in the `saleor/plugins/` directory of the core.
```
korycins, line 9: Don't you think that we could also include some notes about `PaymentGateway` here?  They use the `plugin` interface but they are located in a different directory.

patrys: Good point, done.

docs/developer/extending/plugins.mdx
```diff
@@ -109,14 +104,24 @@ There is no need to implement all base methods as the `PluginsManager` will use
 
 ### Loading your plugin
 
-To activate the new plugin, add it to the `settings.PLUGINS` list in your Django settings:
+Saleor uses the *entry points* API of Python's `setuptools` to automatically discover installed plugins. To make a plugin discoverable, include a `saleor.plugins` entry point in your `setup()` call. The syntax is `package_name = package_name.path.to:PluginClass`:
 
 ```python
-# settings.py
-
-PLUGINS = ["saleor.plugins.custom.plugin.CustomPlugin", ]
```
maarcingebala, line 117: This approach still works but I understand it's only because we have our plugins in Saleor repo and don't recommend that to others building their own plugins?

patrys: It works, in fact discovered plugins are appended to this list. We don't want to recommend modifying core's source code to install a plugin.

PR #137, opened 2020-06-10
Add export products section

IKarbowiak:
Add documentation about exporting products.

### DIFF
--- a/docs/developer.mdx
+++ b/docs/developer.mdx
@@ -23,6 +23,7 @@ Whether you're building a custom-branded storefront or looking for ways to exten
 - [**Checkout**](developer/checkout.mdx): the checkout flow.
 - [**Users**](developer/users.mdx): creating and managing customer accounts.
 - [**Metadata**](developer/metadata.mdx): store additional information about items.
+- [**Importing and exporting**](developer/export-products.mdx): import and export data.
 - [**Extending**](developer/extending.mdx): add custom logic to a Saleor store.
 - [**API Reference**](api-reference.md): the full API reference.
 
--- a/docs/developer.mdx
+++ b/docs/developer.mdx
@@ -23,7 +23,7 @@ Whether you're building a custom-branded storefront or looking for ways to exten
 - [**Checkout**](developer/checkout.mdx): the checkout flow.
 - [**Users**](developer/users.mdx): creating and managing customer accounts.
 - [**Metadata**](developer/metadata.mdx): store additional information about items.
-- [**Importing and exporting**](developer/export-products.mdx): import and export data.
+- [**Exporting data**](developer/export-products.mdx): export data to file.
 - [**Extending**](developer/extending.mdx): add custom logic to a Saleor store.
 - [**API Reference**](api-reference.md): the full API reference.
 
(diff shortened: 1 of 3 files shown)


docs/developer/export-products.mdx
```diff
@@ -0,0 +1,353 @@
...
+(see the `exportInfo` input fields). If no variant fields are provided,
+each row will contain only product fields:
+
+| id  | name         | description                    | category    | product weight |
+|-----|--------------|--------------------------------|-------------|----------------|
+| 61  | Orange Juice | No additives, no preservatives | juices      | 500 g          |
+| 65  | Apple Juice  | No additives, great taste      | juices      | 1000 g         |
+| 101 | Polo Shirt   | Nice and comfortable           | polo-shirts | 200 g          |
+
+### The charts below explains workflow of the exporting products:
```
maarcingebala, line 48: One last suggestion - titles should be concise and rather short, so I would move this sentence to a paragraph and change the title.

```suggestion
### Workflow

The charts below explain the internal workflow of exporting products:
```

docs/developer/export-products.mdx
```diff
@@ -0,0 +1,353 @@
+---
+title: Exporting Products Information
```
maarcingebala, line 2: I think just "Exporting Products" should be fine here:

```suggestion
title: Exporting Products
```

docs/developer.mdx
```diff
@@ -23,6 +23,7 @@ Whether you're building a custom-branded storefront or looking for ways to exten
 - [**Checkout**](developer/checkout.mdx): the checkout flow.
 - [**Users**](developer/users.mdx): creating and managing customer accounts.
 - [**Metadata**](developer/metadata.mdx): store additional information about items.
+- [**Importing and exporting**](developer/export-products.mdx): import and export data.
```
patrys, line 26: It says import and export but only one direction is currently supported.

docs/developer/export-products.mdx
```diff
@@ -0,0 +1,355 @@
+---
+title: Exporting Products
+sidebar_label: Exporting products
```
patrys, line 3: This is only required if the sidebar label (what appears in the sidebar) is different from the title (what appears at the top of the page).

docs/developer/export-products.mdx
```diff
@@ -0,0 +1,355 @@
...
+:::
+
+### Export file structure
+Each row in the exported file represents a single product variant, but it also contains
+general product fields.
+For example, if a product has three variants, there will be three lines in total.
+Each line will contain common product fields such as name or description,
+and fields specific to given variant:
+
+| id | name         | description                    | variant sku | variant weight |
```
patrys, line 32: Are the column titles actually lowercase?

IKarbowiak: Yes

docs/developer.mdx
IKarbowiak: That's true. So maybe, for now, it should be just exporting.

patrys: Yeah, let's call it "exporting data".

PR #143, opened 2020-06-22
Update developer payment-gateways section

IKarbowiak:
Update developer payment-gateways section - some part of this section was out-of-date.


patrys:
Should I backport this to 2.10 docs as well?

PR #144, opened 2020-06-23
Add sorters section

IKarbowiak:
Add document how to add sorter to query resolver and how to customize it.

### DIFF
--- a/docs/developer/extending/api/sorters.mdx
+++ b/docs/developer/extending/api/sorters.mdx
@@ -7,13 +7,13 @@ title: Sorters
 Sorters allow returning elements that are ordered based on specific criteria.
 This document describes how to add a sorter to the query resolver.
 
-## Add sorting to query
+## Add sorter to query resolver
 
 ### Step 1: Create new sort field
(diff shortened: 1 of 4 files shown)


docs/developer/extending/api/sorters.mdx
```diff
@@ -0,0 +1,117 @@
...
+            return f"Sort Example by {sort_name}."
+        raise ValueError("Unsupported enum value: %s" % self.value)
+
+    @staticmethod
+    def qs_with_item_count(queryset: Queryset) -> Queryset:
+        return queryset.annotate(item_count=Count("item"))
+```
+
+### Step 2: Use the field in sort field definition
+The next step is just using it in sort field definition as a normal model field.
```
fowczarek, line 99: Maby we should add some description of that value from (`ITEM_COUNT = ["item_count"]`) this list should be the same as an annotated field name?

IKarbowiak: I added notes in this and the previous section.

PR #175, opened 2020-09-11
Update product API reference fields

IKarbowiak:
Update product API reference with new fields:
- availableForPurchase
- isAvailableForPurchase 
- visibleInListings

### DIFF
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -22881,6 +22881,33 @@ Whether the product is in stock and visible or not.
 </td>
 </tr>
 <tr>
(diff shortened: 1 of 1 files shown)


patrys (approved):

fowczarek (approved):
Maybe we should add a more complex description of `visibleInListings`?
In dashboard we have this description:
```
Disabling this checkbox will remove product from search and category pages. It will be available on collection pages.
```

patrys:
This file is generated, if we need better descriptions, we should fix them in the schema.

PR #191, opened 2020-10-19
Update the payment structures

korycins:


maarcingebala (approved):

docs/developer/extending/payment-gateways.mdx
```diff
@@ -103,18 +106,22 @@ For more information on how to configure your plugin, see the [Plugins](dashboar
...
+| `kind`                          | `str`            | Transaction kind, one of: auth, capture, charge, refund, void            |
+| `amount`                        | `Decimal`        | Amount that the gateway actually charged or authorized                   |
+| `currency`                      | `str`            | Currency in which the gateway charged, needs to be an ISO 4217 code      |
+| `transaction_id`                | `str`            | Transaction ID as returned by the gateway                                |
+| `error`                         | `str`            | An error message, if one occurred. Should be `None` if no error occurred |
+| `customer_id`                   | `str`            | Customer ID                                                              |
+| `card_info`                     | `CreditCardInfo` | Credit Card information                                                  |
+| `raw_response`                  | `dict`           | Raw gateway response as a dict object. By default, it is `None`          |
+| `action_required_data`          | `dict`           | The additional data required to proceed the additional action            |
```
maarcingebala, line 121: To make it easier I would say what can be the example action. When someone is not that familiar with payments they will have no idea what action we mean here.

korycins: Hhm, this is the field used to pass data from payment provider to the storefront, when payment provider requires some action from the user. Based on the Adyen implementation:  the field is fulfilled by the data which needs to be passed to the Adyen's dropin on the storefront
Not sure what exacly I could put this as an example.

patrys: I think we can at least improve the description so it's clear that it contains PSP vendor-specific data

korycins: Will `PSP vendor-specific data,  required to finalize the payment on the frontend` be fine? Or do you expect something more precise/specific? 
@patrys  @maarcingebala

maarcingebala: I think it's better, but would the readers know what is PSP?

maarcingebala: Adyen puts it nicely:

> Some payment methods require additional action from the shopper such as: to scan a QR code, to authenticate a payment with 3D Secure, or to log in to their bank's website to complete the payment.

patrys (changes requested):
Waiting for updated copy

patrys (approved):

PR #196, opened 2020-11-04
Update api reference and metadata for page types

IKarbowiak:
* Add all page type related stuff to api reference
* Update metadata with `page` and `pageType`


IKarbowiak:
History was broken.

PR #239, opened 2021-03-05
Document stock availability fields

maarcingebala:


docs/developer/products.mdx
```diff
@@ -395,6 +395,65 @@ The tax equals 1.15 which is 23% of the net price. Gross price is the sum of net
...
+}
+```
+
+The `address` parameter is used to check the quantity in a warehouse that is connected with a shipping zone that includes the given address. If not provided, Saleor will use a warehouse with the highest available quantity.
+
+The `isAvailable` field is `True` when:
+- there is at least one variant for which the stock quantity minus the allocated quantity is above zero (in a warehouse matched for the given address)
+- the product is published in the given channel and available for purchase (see [Make your product available](/dashboard/catalog/products.md#step-5-make-your-product-available) section in the Dashboard docs)
+
+To fetch the exact available quantity, use the `ProductVariant.quantityAvailable` field:
```
patrys, line 432: I think we should document that quantity available is capped at the maximum orderable threshold to avoid leaking precise stock information.

maarcingebala: Yep, added a note about this.

maarcingebala:
BTW in the light mode, a note has the same background color as inline code formatted with backticks. Maybe we could change it somehow for better readability. But I see that Docuzaurs's docs have the same problem.

See `quantityAvailable` here:

image

docs/developer/products.mdx
```diff
@@ -392,7 +392,70 @@ Result:
...
 :::note
-Note: The `address` parameter is optional. If it's not provided, Saleor will use the default country code configured in the backend settings (see [DEFAULT_COUNTRY](running-saleor/configuration.mdx#default_country)).
+The `address` parameter is optional. If it's not provided, Saleor will use the default country code configured in the backend settings (see [DEFAULT_COUNTRY](running-saleor/configuration.mdx#default_country)).
+:::
+
+## Stock availability
+
+Stock availability of products is defined for each variant and warehouse (see [Warehouses](/dashboard/configuration/warehouses.mdx) and [Checking inventory](/dashboard/catalog/products.md#checking-inventory) sections to see how to configure it in the Dashboard). There are two fields that allows you to fetch the stock information in the public API.
+
+To fetch the stock availability, use the `Product.isAvailable` field:
```
krzysztofwolski, line 402: https://docs.saleor.io/docs/next/developer/api-reference/objects/product#isavailable-boolean
Please update API reference (instruction in the readme).

krzysztofwolski (changes requested):

docs/developer/products.mdx
maarcingebala: Update API reference about stock related changes.

PR #242, opened 2021-03-17
Freeze docusaurus versions

krzysztofwolski:
Previous build was failing because of breaking changes from new docusaurus version. In this PR I'm freezing its reelease to avoid such situations in thee future.


.github/workflows/main.yml
```diff
@@ -12,6 +12,8 @@ jobs:
     - uses: actions/checkout@v2
     
     - uses: actions/setup-node@v1.4.4
+      with:
+        node-version: '12'
```
patrys, line 16: I'm 👎. This is going to collect dust, especially given that the current version is 14 already.

website/package.json
```diff
@@ -11,11 +11,11 @@
     "graphql-markdown": "^5.2.0"
   },
   "dependencies": {
-    "docusaurus2-graphql-doc-generator": "file:docusaurus2-graphql-doc-generator",
-    "@docusaurus/core": "^2.0.0-alpha.70",
-    "@docusaurus/preset-classic": "^2.0.0-alpha.70",
-    "@docusaurus/types": "^2.0.0-alpha.70",
+    "@docusaurus/core": "==2.0.0-alpha.70",
+    "@docusaurus/preset-classic": "==2.0.0-alpha.70",
+    "@docusaurus/types": "==2.0.0-alpha.70",
```
patrys, line 16: Why? There's a lock file.

krzysztofwolski: with `^` pre-release versions were updating in lock. And those had breaking changes :/ 
Anyway, cant find issue why those builds are failing on CI, tried to clean my node_modules and create lock again, and no luck with that. Node version freeze was blind try to find whats different about my local env and CI

patrys: The lock file does not update itself unless you explicitly tell npm to upgrade a dependency (in which case you are responsible for verifying its compatibility).

PR #248, opened 2021-03-22
Add S3 ACL description

### LINKED ISSUE #34: S3 ACL description in deployment guide
Settings.py: `AWS_DEFAULT_ACL = os.environ.get("AWS_DEFAULT_ACL", None)`

By default uploaded files will have permissions inherited from s3 bucket settings. It may be a good idea to write about this in docs. 
In our case it was confusing because images uploaded from dashboard were not accessible by the users. In our case we decided to use policy `public-read)`

krzysztofwolski:
Fixes #34

Misconfiguring access permissions is a common issue with deployments. Added short description and linked AWS docs.

### DIFF
--- a/docs/developer/running-saleor/s3.mdx
+++ b/docs/developer/running-saleor/s3.mdx
@@ -11,13 +11,14 @@ This integration allows you to delegate storing such files to [Amazon’s S3 ser
 | Variable Name              | Description                                     |
 |----------------------------|-------------------------------------------------|
 | `AWS_ACCESS_KEY_ID`        | Your AWS access key.                            |
-| `AWS_SECRET_ACCESS_KEY`    | Your AWS secret access key.                     |
+| `AWS_DEFAULT_ACL`          | Default [Access Control List](https://docs.aws.amazon.com/AmazonS3/latest/userguide/acl-overview.html) to uploaded files. |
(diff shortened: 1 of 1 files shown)


docs/developer/running-saleor/s3.mdx
```diff
@@ -11,13 +11,14 @@ This integration allows you to delegate storing such files to [Amazon’s S3 ser
 | Variable Name              | Description                                     |
 |----------------------------|-------------------------------------------------|
 | `AWS_ACCESS_KEY_ID`        | Your AWS access key.                            |
-| `AWS_SECRET_ACCESS_KEY`    | Your AWS secret access key.                     |
+| `AWS_DEFAULT_ACL`          | Default [Access Control List](https://docs.aws.amazon.com/AmazonS3/latest/userguide/acl-overview.html) to uploaded files. |
```
NyanKiyoshi, line 14: When reading this I had a concern: "Default to what? Is Saleor trying to do something I did not ask it to: setting the ACL?" - can you mention (if my guess is right) that if none is provided, it will be leveraged to the s3 settings? (meaning that Saleor does not try to guess anything, pushes the file without setting any ACL)

krzysztofwolski: Good feedback, thanks!

NyanKiyoshi (approved):

karolkielecki (approved):

PR #258, opened 2021-04-22
Clarify stock quantities query

krzysztofwolski:
Based on developer feedback, I would like to clarify how quantities behave in API

### DIFF
--- a/docs/developer/products.mdx
+++ b/docs/developer/products.mdx
@@ -395,7 +395,7 @@ The tax equals 1.15 which is 23% of the net price. Gross price is the sum of net
 The `address` parameter is optional. If it's not provided, Saleor will use the default country code configured in the backend settings (see [DEFAULT_COUNTRY](running-saleor/configuration.mdx#default_country)).
 :::
 
-## Stock availability
+## Stock availability for customers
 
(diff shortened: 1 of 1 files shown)


fowczarek (approved):

docs/developer/products.mdx
```diff
@@ -452,12 +452,68 @@ Response:
...
 
+## Stock quantities
+
+For creating integrations with external stock management system, you will need stock values not restricted by `MAX_CHECKOUT_LINE_QUANTITY`. 
+
+:::note
+`stocks` field in the [`ProductVariant`](developer/api-reference/objects/product-variant.mdx) object require `MANAGE_PRODUCTS` permission.
+:::
+
+To clarify difference between `quantityAvailable` and stocks, we will query both for the product variant:
```
karolkielecki, line 469:
```suggestion
To clarify the difference between `quantityAvailable` and stocks, we will query both for the product variant: 
```

docs/developer/products.mdx
```diff
@@ -452,12 +452,68 @@ Response:
 }
 ```
 
-The `address` parameter behaves in the same way as in the `Product.isAvailable` fields above.
+Warehouse is chosen based on argument `address`. If argument is ommited, API will return highest available quantity.
```
karolkielecki, line 455:
```suggestion
A warehouse is chosen based on the argument `address`. If the argument is omitted, API will return the highest available quantity.
```

karolkielecki (changes requested):

PR #263, opened 2021-05-10
Add new variant logic

karolkielecki:


krzysztofwolski (approved):

docs/dashboard/catalog/products.md
```diff
@@ -211,7 +211,9 @@ Categorize your new product using the drop-down menus to assign a relevant Produ
 
 Once you have set up all your product details, you need to assign a previously created [channel](dashboard/configuration/channels.md). Click the _MANAGE_ button in the top right corner of the availability card, and then assign a channel. You can assign as many channels as you need for each product.
 
-Each channel gives you fine control over the availability of individual items. You can set a product's publication date on a per-channel basis.
+Each channel gives you fine control over the availability of individual items. You can choose all or specific variants that will be assigned to a channel. Then you can set a product's publication date on a per-channel basis.
+
+![](../screenshots/catalog-product-variant-availability.png)
```
maarcingebala, line 216: The screenshot doesn't look too good for two reasons:
- scrollbars
- channel name is "NightCity" - is this a good example for the docs? I would say something neutral like "Default Channel" or "Channel USD"

What browser do you use to take screenshots? I'm using Chrome on Mac and I don't see these scrollbars:
image

docs/dashboard/catalog/products.md
```diff
@@ -366,6 +366,10 @@ To track inventory, you need to enable the _Track Inventory_ checkbox.
 
 You can repeat this process for all sizes or types of the same product by clicking the Variants card on the left side. Each variant has a specific SKU code (Stock Keep Unit) which is unique to specific variants.
 
+Once the product availability is set on the main product page, you will also find information as to which channel a specific variant of a product is assigned to at the top of each variant page.
+
+![Editing product variants](../screenshots/catalog-product-variant-details.png)
```
maarcingebala, line 371: This screenshot should be updated as well but I think we're using the old one here? A new section was added:
image

docs/dashboard/catalog/products.md
karolkielecki: I'm using Chrome on Windows. Yes, we've used "NightCity" in other screenshots and assets as an easter egg--I can change it.

karolkielecki: @maarcingebala It's a bug. There is already a ticket for it, but as far as I know, it has a low priority.

maarcingebala: While the bug is waiting for a fix, can't we use a screenshot made on Mac?

docs/dashboard/catalog/products.md
karolkielecki: @maarcingebala I'm working on the master staging and that's the only new section I see on the top of the variant page.

PR #264, opened 2021-05-17
Add required permissions section to the checkouts

krzysztofwolski:


korycins (approved):

docs/developer/checkout.mdx
```diff
@@ -21,6 +21,10 @@ Depending on the chosen channel, the user will have access to different objects.
 
 [Learn more about using multiple channels](developer/channels.mdx).
 
+## Permissions
+
+If Checkout is assigned to the user, only this user and staff user with permission `MANAGE_CHECKOUTS` can query its data. Checkouts without assigned user can be queried modified using token, without additional permissions.
```
karolkielecki:
```suggestion
If the checkout permission is assigned to the user, only this user and staff users with permission `MANAGE_CHECKOUTS` can query its data. Checkouts without assigned users can be queried modified using the token without additional permissions.  
```

karolkielecki (changes requested):

karolkielecki (approved):

Issue #100, opened 2020-02-07
Document OpenTracing configuration


maarcingebala:
We've recently added OpenTracing support. We need to add documentation to describe:
- what is OpenTracing and what it does (briefly)
- how to configure it in Saleor
- example of how to use it with [Jaeger](https://www.jaegertracing.io/)

krzysztofwolski:
@maarcingebala https://docs.saleor.io/docs/3.0/developer/running-saleor/monitoring#jaeger
Does this page needs improvement?

maarcingebala:
The part about Jaeger looks up to date.

PR #281, opened 2021-06-21
Add mailhog and ngrok descriptions

### LINKED ISSUE #81: Write documentation on exposing the locally running instance
For certain tasks the developer may need to expose a locally running instance of Saleor to the outside world. One such case is payments integration where callbacks happen over web hooks. We should document working with tools such as `ngrok` to make this easier.

krzysztofwolski:
- exposing local env with ngrok
- using mailhog

Close #81


krzysztofwolski:
I'll write open tracing docs in separate PR (ref #100)

krzysztofwolski:
Also - I'll add exposing localhost info to the apps section in the separate PR

docs/developer/running-saleor/exposing-instance.mdx
```diff
@@ -0,0 +1,20 @@
+---
+title: Exposing local instance
+sidebar_label: Exposing instance
+---
+
+There are situations when you would like to expose your local environment to external clients, for example:
+
+- show your project to the manager.
+- integrate with the new payment gateway, which requires a return URL.
```
korycins, line 9: This won't work. There is a settings.ALLOWED_CLIENT_HOSTS which defines where user can be redirected. If you don't overwrite this variable, the allowed redirect URL will be 127.0.0.1 and localhost. Any different than this will raise an error. The same case is for the situation when you're hosting storefront somewhere, and you use ngrok to connect it to your backend (all mutation like, reset password, send activation URL will raise an exception, afair).

docs/developer/running-saleor/debugging-emails.mdx
```diff
@@ -0,0 +1,28 @@
+---
+title: Debugging emails
+sidebar_label: Debugging emails
+---
+
+[Saleor platform](https://github.com/mirumee/saleor-platform) use [Mailhog](https://github.com/mailhog/MailHog) for catching all the emails send by the Saleor. 
+
+![Assign product to collection](../screenshots/mailhog.png)
+
+Pros of using Mailhog during development:
```
korycins, line 10: IMHO this chapter should convince user to configure UserEmailPlugin(or AdminEmailPlugin), instead of changeing env settings. Each of these actions can be done directly from the Dashboard's plugin configuration view. Additionally, you can upload there your own email templates, which will allow you to also debug email content (with described approach, we can only debug a communication Saleor -> Emailgatewat(mailhog)).

karolkielecki (changes requested):

docs/developer/running-saleor/debugging-emails.mdx
```diff
@@ -0,0 +1,28 @@
+---
+title: Debugging emails
+sidebar_label: Debugging emails
+---
+
+[Saleor platform](https://github.com/mirumee/saleor-platform) use [Mailhog](https://github.com/mailhog/MailHog) for catching all the emails send by the Saleor. 
+
+![Assign product to collection](../screenshots/mailhog.png)
```
NyanKiyoshi, line 8: Alternative text does not seem to match the screenshot

docs/developer/running-saleor/debugging-emails.mdx
```diff
@@ -0,0 +1,28 @@
...
+
+Pros of using Mailhog during development:
+- live preview of rendered emails.
+- no need to send emails to external services.
+- test emails will be caught in the local environment, so there is no possibility of sending test data to actual customers.
```
NyanKiyoshi, line 13: Possibly, we could add that it also prevent hurting the mail reputation due to bounces. e.g. sending emails to `foo@example.com`, which will cause the email to bounce and thus lower the reputation

docs/developer/running-saleor/exposing-instance.mdx
```diff
@@ -0,0 +1,21 @@
...
+- testing app deployed in the cloud.
+
+To achieve that, you need a service for redirecting the traffic. Such service delivers [ngrok](https://ngrok.com) and [localtunnel](https://localtunnel.me). Both are great for our use case, and you can choose any of those services.
```
karolkielecki, line 12:
```suggestion
To achieve that, you need a service for redirecting the traffic. Such service is delivered by [ngrok](https://ngrok.com) and [localtunnel](https://localtunnel.me). Both are great for our use case, and you can choose any of those services.
```

karolkielecki (changes requested):

karolkielecki (approved):

PR #287, opened 2021-07-01
Bump dependencies

krzysztofwolski:


patrys:
Did you check if the charts continue to work and look correct? I've seen them break with previous mermaid upgrades.

krzysztofwolski:
@patrys Checked charts and it seemed okay (at least the apps section). I'm kinda puzzled with CI fail - everything worked on the local machine, and docusaurus issues only mentioned some yarn issues. So far update is not crutial - I'll come back to this issue after a higher priority task. 
Hoped that this update will be quick job, but it doesn't seems to be the case 😭

krzysztofwolski:
Looks like it needed some time to bloom. All green, merging.
Tested styles on local env and all components looks good.

PR #300, opened 2021-08-06
Update API reference

maarcingebala:
Update API ref and mention new translation webhooks.


krzysztofwolski (changes requested):
@maarcingebala could you fix those descriptions in the API? Space is missing before the `DEPRECATED`

maarcingebala:
@krzysztofwolski I'll fix them later but I guess it's not a blocker for updating the API ref; it's a separate issue in the core repo and API ref will be automatically fixed with the next update after we fix the issue in the core.

krzysztofwolski (approved):

krzysztofwolski:
Ok, lets do this this way

PR #303, opened 2021-08-16
Update structure and make v3.0 default

krzysztofwolski:
- moved contents of `/website` to the root dir
- changed the latest (default) version to the 3.0
- now all URLs will contain a version number, even the "current master"
    - if the URL does not contain the version number, the 404 page will try to append 3.0 to it
- docusaurus version bump


src/pages/versions.js
```diff
@@ -44,7 +44,7 @@ function Versions() {
               <tr>
                 <th>{latestStableVersion}</th>
                 <td>
-                  <Link to={useBaseUrl("/docs/")}>Documentation</Link>
+                  <Link to={useBaseUrl("/docs/2.11/")}>Documentation</Link>
```
patrys, line 47: Why hardcode the version string here if `latestStableVersion` is calculated?

patrys (approved):
I'd consider adding a comment next to the 2.11 redirect callback to explain what it does and that we should eventually remove it.

patrys:
Task linked: [SALEOR-4262 Make 3.0 default version of the docs](https://app.clickup.com/t/2549495/SALEOR-4262)

PR #332, opened 2021-11-02
Update docs for metadata - add note that token can be used for modyfing checkout and order metadata

IKarbowiak:
Updates for https://github.com/saleor/saleor/pull/8426

### DIFF
--- a/docs/developer/metadata.mdx
+++ b/docs/developer/metadata.mdx
@@ -113,6 +113,13 @@ type Mutation {
 }
 ```
 
+:::note
+For `Order` and `Checkout` types it is recommended to use `token` as the `id` input value,
+for other types use GraphQL ID.
+
+Changing the `Order` metadata is possible only with the use of `token`.
+:::
+
 All resulting types have the same shape:
 
 ```graphql


docs/developer/metadata.mdx
```diff
@@ -113,6 +113,15 @@ type Mutation {
 }
 ```
 
+:::note
+For `Order` and `Checkout` as the `id` input value, you can provide either `token` or `id` value.
+:::
+
+:::warning
+In the near future, changing `Order` metadata will be available only with the use of `token`.
```
L3str4nge, line 121: How about to specify version number?

L3str4nge (approved):

docs/developer/metadata.mdx
IKarbowiak: I can add. @maarcingebala which version?

maarcingebala: We want to do that before release stable 3.0, and the docs are targeting the 3.0 version. Maybe we don't have to mention that it will change, but at this point we should say that with order and checkout type you should use token as identifier, while for other types you should use GraphQL ID?

docs/developer/metadata.mdx
```diff
@@ -113,6 +113,13 @@ type Mutation {
 }
 ```
 
+:::note
+For `Order` and `Checkout` is recommended to use `token` as the `id` input value,
```
maarcingebala, line 117:
```suggestion
For `Order` and `Checkout` types it is recommended to use `token` as the `id` input value,
```

docs/developer/metadata.mdx
```diff
@@ -113,6 +113,13 @@ type Mutation {
...
+for other types use GraphQL ID.
+
+Changing `Order` is possible only with the use of `token`.
```
maarcingebala, line 120:
```suggestion
Changing the `Order` metadata is possible only with the use of `token`.
```

maarcingebala (approved):

PR #392, opened 2022-03-23
Add new collection screenshots

karolkielecki:


Qasem-h (approved):

krzysztofwolski (changes requested):
Waiting for the fix on the frontend side to remove unnecessary scrollbars

peelar:
Hi @karolkielecki 👋 

What is the status here? Do we still need it?

karolkielecki:
@peelar IMHO, if we change our docs structure and approach (less or no screenshots) -- then I think it's not needed.

PR #443, opened 2022-06-21
Add husky with prettier

timuric:
Run prettier on changed/staged files. https://prettier.io/docs/en/precommit.html#option-2-pretty-quickhttpsgithubcomazzpretty-quick

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,12 +17,15 @@
     "publish-gh-pages": "docusaurus deploy",
     "version": "docusaurus version",
     "rename-version": "docusaurus rename-version",
-    "update-api-reference": "docusaurus graphql-to-doc --force --pretty"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -17,12 +17,15 @@
     "publish-gh-pages": "docusaurus deploy",
     "version": "docusaurus version",
     "rename-version": "docusaurus rename-version",
-    "update-api-reference": "docusaurus graphql-to-doc --force --pretty"
+    "update-api-reference": "docusaurus graphql-to-doc --force --pretty",
+    "prepare": "husky install"
```
bmigirl, line 21: as far as i understand this should only be run once so we can remove it I guess, but where is husky dir with husky commands?

timuric: @bmigirl you mean prepare script?

witoszekdev (approved):

PR #445, opened 2022-06-21
Add the webhook verification code example for NodeJS.

nnthuan:


peelar:
Hello @nnthuan 👋 

It is very flattering that you spend your time trying to improve our documentation. We are beyond grateful for that ❤️ .

At the same time, we have to be cautious about what we add to our docs. Every piece of code is something we have to maintain, and we could not promise that with the snippet you supplied.

We also have [our own version of the webhook payload validation](https://github.com/saleor/saleor-app-sdk/blob/main/src/middleware.ts#L59), and wherever possible, we would like to stick to using this one.

Thank you for your effort to improve the experience of using Saleor 🙏, looking forward to seeing more from you.

PR #456, opened 2022-07-01
Add documentation about graphql tools

timuric:


docs/developer/getting-started/playground.mdx
```diff
@@ -14,7 +16,18 @@ To view and try out the Playground instance of Saleor's demo store, visit the [S
 
 For more information, visit the [Playground GitHub page](https://github.com/prisma/graphql-playground).
 
-
-## Authentication
+#### Authentication
```
patrys, line 19: Shouldn't it be a third level header? Won't making it fourth level remove it from the outline?

patrys (approved):
See my comment about header levels, other than that LGTM. :shipit:

docs/developer/getting-started/playground.mdx
timuric: I understand the semantics, I tried that, but it has low contrast with the header bellow, it feels competing (we should use higher values for typography scale). I can change it if you think visually it doesn’t make sense

PR #463, opened 2022-07-14
Add docs for better media thumbnails

IKarbowiak:
Add a `thumbnails` section.

### DIFF
--- a/docs/developer/thumbnails.mdx
+++ b/docs/developer/thumbnails.mdx
@@ -1,0 +1,92 @@
+---
+title: Thumbnails
+---
+
+## Introduction
+This guide describes how to ask for a thumbnail and how the thumbnails are generated.
+
+
+### Thumbnails sizes and formats
+Thumbnails are available on the following models and fields:
+- `User.avatar`
+- `OrderLine.thumbnail`
+- `Product.thumbnail`
+- `Collection.background_image`
+- `Category.background_image`
+- `ProductMedia.url`
+- `ProductImage.url`
+
+You can specify the `size` and `format` of the thumbnail. The size can be any integer value,
+but the resolution with the closest value from the following list will be used:
+[32, 64, 128, 256, 512, 1024, 2048, 4096].
+The format value must be provided together with the size value. Right now only `WebP` format
+can be specified, when not provided the thumbnail with the original format will be returned.
+Either size and format are not required, when these parameters are not provided,
+the original image is returned.
+
+### Generating the thumbnails
+The thumbnails are generated on-demand. When the user is asking for the thumbnail with
+given size and format for the first time, the proxy URL is returned. When the proxy URL
+is opened, the requested thumbnail is created and the user get the requested image.
+In case of a second and further request for the same thumbnail, the previously created
+thumbnail's URL is returned.
+
+### Example
+Here is an example of query the categories background image of size 100 and format WebP
+for the first time:
+```graphql
+query {
+    category(id: "Q2F0ZWdvcnk6MjU=") {
+        name
+        backgroundImage(size: 100, format: WEBP) {
+            url
+            alt
+        }
+    }
+}
+```
+
+In the response we get the proxy URL.
+```json
+{
+	"data": {
+		"category": {
+			"name": "Accessories",
+			"backgroundImage": {
+				"url": "http://localhost:8000/thumbnail/Q2F0ZWdvcnk6MjU=/128/webp/",
+				"alt": ""
+			}
+		}
+	},
+	"extensions": {
+		"cost": {
+			"requestedQueryCost": 1,
+			"maximumAvailable": 50000
+		}
+	}
+}
+```
+
+After opening the link, the thumbnail with the size 128 and WebP format is generated.
+The resolution 128 is used as this is the closest available size to the provided value. 
+
+The next time we are asking for the same thumbnail we are getting the URL to the actual thumbnail.
+```json
+{
+	"data": {
+		"category": {
+			"name": "Accessories",
+			"backgroundImage": {
+				"url": "http://localhost:8000/media/thumbnails/category-backgrounds/20210809_181533_2240b122_thumbnail_128.webp",
+				"alt": ""
+			}
+		}
+	},
+	"extensions": {
+		"cost": {
+			"requestedQueryCost": 1,
+			"maximumAvailable": 50000
+		}
+	}
+}
+```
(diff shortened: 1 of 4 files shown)


maarcingebala (approved):

docs/developer/thumbnails.mdx
```diff
@@ -0,0 +1,92 @@
...
+- `ProductMedia.url`
+- `ProductImage.url`
+
+You can specify the `size` and `format` of the thumbnail. The size can be any integer value,
+but the resolution with the closest value from the following list will be used:
+[32, 64, 128, 256, 512, 1024, 2048, 4096].
+The format value must be provided together with the size value. Right now only `WebP` format
+can be specified, when not provided the thumbnail with the original format will be returned.
+Either size and format are not required, when these parameters are not provided,
+the original image is returned.
```
peelar, line 25:
```suggestion
Currently, `WebP` is the only supported value for `format`.

Both `size` and `format` fields are optional, but when you decide to provide `format`, the `size` becomes required as well. If either, or both of these are not specified, the original image attributes will be used.
```

docs/developer/thumbnails.mdx
```diff
@@ -0,0 +1,92 @@
...
+### Generating the thumbnails
+The thumbnails are generated on-demand. When the user is asking for the thumbnail with
+given size and format for the first time, the proxy URL is returned. When the proxy URL
+is opened, the requested thumbnail is created and the user gets the requested image.
+In case of a second and further request for the same thumbnail, the previously created
+thumbnail's URL is returned.
+
+### Example
+Here is an example of query the categories background image of size 100 and format WebP
+for the first time:
```
peelar, line 36:
```suggestion
Here is an example query for the category's background image of `size` = "100" and `format` = "WebP":
```

docs/developer/thumbnails.mdx
```diff
@@ -0,0 +1,92 @@
...
+        name
+        backgroundImage(size: 100, format: WEBP) {
+            url
+            alt
+        }
+    }
+}
+```
+
+In the response we get the proxy URL.
```
peelar, line 49:
```suggestion
As the query ran for the first time, the response contains proxy URL:
```

docs/developer/thumbnails.mdx
```diff
@@ -0,0 +1,92 @@
...
+		"cost": {
+			"requestedQueryCost": 1,
+			"maximumAvailable": 50000
+		}
+	}
+}
+```
+
+After opening the link, the thumbnail with the size 128 and WebP format is generated.
+The resolution 128 is used as this is the closest available size to the provided value.
```
peelar, line 71:
```suggestion
After opening the link, the thumbnail with the `size` of "128", and the `format` of "WebP" is generated.
The resolution is equal to 128 as this is the closest available size to the provided value. 
```

docs/developer/thumbnails.mdx
```diff
@@ -0,0 +1,92 @@
...
+			"maximumAvailable": 50000
+		}
+	}
+}
+```
+
+After opening the link, the thumbnail with the size 128 and WebP format is generated.
+The resolution 128 is used as this is the closest available size to the provided value. 
+
+The next time we are asking for the same thumbnail we are getting the URL to the actual thumbnail.
```
peelar, line 73:
```suggestion
The next time we run the query, we receive the URL to the actual thumbnail.
```

docs/developer/thumbnails.mdx
```diff
@@ -0,0 +1,100 @@
...
+    name
+    backgroundImage(size: 100, format: WEBP) {
+      url
+      alt
+    }
+  }
+}
+```
+
+As the query ran for the first time, the response contains proxy URL:
```
pawelpikus, line 55: You mean: "After the query runs ..., it contains the proxy?"
"As" introduces two actions happening at the same time, so here it would mean that the response contains URL **in the middle of** running the query. What we want to say is: The query runs for the first time and as a result the response contains proxy URL, right? Or am I wrong?

pawelpikus (commented):
Hello👋
Looking good already. I just focused on getting rid of the passive voice, whenever we want to focus on the user doing the action and not so much on the action itself. Also, there's one thing that needs clarification.

Hope it helps!

docs/developer/thumbnails.mdx
peelar: That's correct. It was more like "because the query ran for the first time, the response looks like this. It would look differently, if it would run again"

peelar:
> I just focused on getting rid of the passive voice, whenever we want to focus on the user doing the action and not so much on the action itself

That's such a great insight @pawelpikus! I am grateful to be able to draw from your expertise in these topics 🙌.

IKarbowiak:
Thank you both guys, @peelar @pawelpikus! It sounds much better after your suggestions. 🚀

peelar (approved):

docs/developer/thumbnails.mdx
```diff
@@ -0,0 +1,100 @@
...
+
+You can specify the `size` and `format` of the thumbnail. You are free to pass any integer
+as `size`, but what ends up being used is the closest value from the following list:
+[32, 64, 128, 256, 512, 1024, 2048, 4096].
+For example: when you pass "55", we resolve it to "64".
+
+Currently, `WebP` is the only supported value for `format`.
+Both `size` and `format` fields are optional, but when you decide to provide `format`,
+the `size` becomes required as well. If either, or both of these are not specified,
+the original image attributes will be used.
```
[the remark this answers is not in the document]
timuric: Does it mean that I can provide `( size )` or `(format, size)` but not `(format)` @pawelpikus @IKarbowiak?

I feel that this sentence could be more explicit and succint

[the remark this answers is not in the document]
timuric: @CossackDex could you please check in code if this is the case?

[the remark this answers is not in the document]
timuric: So if I undestand correctly the description should be:

Both `size` and `format` arguments are optional with fallback to original image attributes. Argument `format` should be passed together with `size`.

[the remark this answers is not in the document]
peelar: It's:
- nothing
- just `size`
- `format` and `size`

There can't be just `format`.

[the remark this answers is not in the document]
timuric: @pawelpikus please review my coment above

PR #468, opened 2022-07-20
Add stock allocation section

IKarbowiak:
- Create `Stock allocation` section. 
- Move the `shipping zone` section to `Stock allocation` section.

☝️ Small section regarding c&c will be added in another PR.


peelar (commented):
**Question**: I tried to search for the section you added @IKarbowiak through the Algolia search in the top right corner, but it would not display there. What's the process of populating the data Algolia uses?

@timuric

docs/developer/stock-allocation.mdx
```diff
@@ -0,0 +1,298 @@
...
+    class Warehouse{
+      click_and_collect_option: String
+    }
+```
+
+## Allocation flow
+
+Stocks are allocated in the checkout completing process, for all lines
+that have track inventory turned on. The allocation is made for every line
+on stocks with the highest quantity available value. When there is not enough
```
peelar, line 52:
```suggestion
that have track inventory turned on. The allocation is made for every line
on stocks with the highest quantity value available. When there is not enough
```

I can be wrong, though, because I am not sure if I understand the context correctly.

docs/developer/stock-allocation.mdx
```diff
@@ -0,0 +1,298 @@
...
+available quantity in one stock, multiple allocations are created. In case there is
+no enough quantity in all product stocks - the `InsufficientStock` error is raised.
+
+The line can be allocated only in the stocks that fulfill the following conditions:
+
+- the stock's warehouse is available in the channel in which the order is created
+- the stock's warehouse has a shipping zone that includes the shipping address
+  (or billing address if the shipping address is not set) country code
+  and this shipping zone is available in the channel in which the order is created
+- the stock has available quantities
```
peelar, line 62: **Comment**:
This is super helpful. In the future, it would be nice to enrich these kinds of descriptions with screenshots.

docs/developer/stock-allocation.mdx
```diff
@@ -0,0 +1,298 @@
...
+
+## Warehouse configuration
+
+### Warehouse creation
+
+Firstly we need to create the warehouse with use of
```
peelar, line 68:
```suggestion
Firstly we need to create the warehouse with the use of
```

docs/developer/stock-allocation.mdx
```diff
@@ -0,0 +1,298 @@
...
+      "errors": []
+    }
+  }
+}
+```
+
+After the warehouse is created it can be changed with the use of
+the [`updateWarehouse`](developer/api-reference/mutations/update-warehouse.mdx) mutation.
+There is an option to change the following warehouse fields: `name`, `slug`, `email`, `address` and
+also define privacy and click and collect options.
```
peelar, line 129:
```suggestion
also define privacy and "click and collect" options.
```

docs/developer/stock-allocation.mdx
```diff
@@ -0,0 +1,298 @@
...
+Shipping zones contain shipping methods that are available in given countries and
+channels.
+
+Specifying shipping zones is required to allow the selling of physical products to given countries.
+
+### Shipping zone creation and update
+
+To create the shipping zone use [`shippingZoneCreate`](developer/api-reference/mutations/shipping-zone-create.mdx) mutation.
+You can specify such fields as `name`, `description`, `countries`, `warehouses`, `channels` and
+decide if the shipping zone should be used as default one for all countries
```
peelar, line 195:
```suggestion
decide if the shipping zone should be used as the default one for all countries
```

peelar (commented):
Looks really informative! I could use it when I was building my shop 😵. 

A part of the bits I pointed out are coming straight from [Grammarly](https://www.grammarly.com/) which I will evangelize using.

The others are a bit more subjective, I am nevertheless open to discussing any of them.

docs/developer/stock-allocation.mdx
IKarbowiak: IMO it shouldn't be changed. I'm writing about the value that is called `available quantity` and it's the same as the `quantity`. Some `quantity` must be allocated or reserved and in such case, both values differs.

peelar: Got it! Thanks for explanation.

docs/developer/stock-allocation.mdx
```diff
@@ -0,0 +1,298 @@
...
+The warehouses, shipping zones, and channels must be properly linked together
+to provide the expected allocation. Below you can find the explanation of the terms
+that will be required to properly understand this guide:
+
+- channel - allow configuring your shop differently for different currencies;
+  go [here](./channels.mdx) for more information
+- shipping zone - allow for defining the cost of shipping in different areas of the world,
+  shipping zone contains shipping methods that can be used for a given part of the world
+- warehouse - contains stocks of specific products, assigned to a given channel and shipping zone,
+  can be defined as the place for click and collect
```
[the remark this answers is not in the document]
maarcingebala: We call it a "collection point" in API descriptions - I would rephrase and use it here as well.

docs/developer/stock-allocation.mdx
maarcingebala: Do you mean dashboard screenshots? This might be a good idea, although dev docs are more API-oriented. We could also link to dashboard sections here.

maarcingebala: Something to discuss probably as it should be unified across other guides (maybe as a separate task).

peelar: > Do you mean dashboard screenshots? This might be a good idea, although dev docs are more API-oriented.

Now that I think about it, I think you are right. As far as I can see, we don't have dashboard screenshots in the "Developer's Guide" right now, and that's because of the clear line between what those two tutorials are. Maybe you are deep in on some technical problems, and rolled your own dashboard, so why are you seeing the screenshots of the Saleor dashboard?

<thinking_out_loud>

I am just thinking about my experience of starting with Saleor: sometimes I got lost in the "Dashboard Manual", so I was looking for answers in the "Developer's Guide". That is not necessarily a reason to add Dashboard screenshots there, but for sure to meticulously reference sections from "Developer's Guide" inside "Dashboard Manual".

</thinking_out_loud>

timuric: @peelar I really like a visual aid, but one thing to keep in mind about screenshots is that they get dated, and become a source of confusion.

maarcingebala (approved):

peelar (approved):

PR #473, opened 2022-07-26
Fix link to point at playground

jmalk:
I think the second link should point at the https://demo.saleor.io/graphql/, rather than the getting started docs.


jmalk:
Not getting a response from my employer about the CLA so unfortunately will just have to close this.

PR #478, opened 2022-07-31
Update Docker Compose installation process

MohamedAliRashad:
Move from `docker-compose` to `docker compose`

### DIFF
--- a/docs/developer/installation.mdx
+++ b/docs/developer/installation.mdx
@@ -32,20 +32,20 @@ Saleor uses shared folders to enable live code reloading. If you're using **Wind
 Apply database migrations, collect static assets:
 
 ```shell
-docker-compose run --rm api python3 manage.py migrate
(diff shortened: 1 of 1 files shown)


peelar:
Hello @MohamedAliRashad!

First of all, much thanks for your willingness to contribute ⭐.

We are keeping a close eye on the topic but we want to sit on our hands for some more time with Docker Compose V2. One of the big reasons for that is the Linux support that is not fully there yet. `docker-compose` works just fine, so it will have to suffice for now.

Have a nice day!

MohamedAliRashad:
@peelar Thanks for your kind replay.

I just want to point this announcement from Docker regarding Docker Compose V1 Deprcation: 
https://www.docker.com/blog/announcing-compose-v2-general-availability/

Feel free to close the PR anytime ^^

PR #493, opened 2022-08-22
refactor/docs structure

peelar:


patrys:
@peelar I did not want to change the file structure as that requires redirects to be configured for all existing URLs (otherwise Google results will lead to a 404 page).

peelar:
> @peelar I did not want to change the file structure as that requires redirects to be configured for all existing URLs (otherwise Google results will lead to a 404 page).

I understand this is inconvenient and we definitely don't want to do it often but isn't a big refactoring of the documentation structure a good opportunity to clean up the paths as well? I can perform it, no problem.

pawelpikus (approved):

timuric (approved):

PR #506, opened 2022-08-24
Add Sentry integration

peelar:
**Note:**
I had to swizzle the `Layout` component because AFAIK, there was no "general wrapper" for the entire app.

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -1,3 +1,4 @@
+require("dotenv").config();
 const path = require("path");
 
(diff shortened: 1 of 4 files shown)


patrys (approved):

patrys:
I think the best approach would be to write a Docusaurus plugin (there was one for Docusaurus 1.x) but for now, this is probably fine.

pawelpikus (approved):

timuric (approved):

PR #510, opened 2022-08-25
Proofreading🔍: Users

pawelpikus:
Mostly grammar and punctuation fixes, a bit of sentence clarification.

### DIFF
--- a/docs/developer/users.mdx
+++ b/docs/developer/users.mdx
@@ -5,7 +5,7 @@ sidebar_label: Users
 
 ## Introduction
 
-This guide describes how to use the Saleor GraphQL API to register a new account as a customer and do common account management operations.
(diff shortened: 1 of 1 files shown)


docs/developer/users.mdx
```diff
@@ -130,7 +130,7 @@ mutation {
 }
 ```
 
-As a result, we get data of the newly created, but yet inactive user:
+As a result, we the get data of the newly created but yet inactive user:
```
peelar, line 133:
```suggestion
As a result, we get data of the newly created but yet inactive user:
```

peelar (approved):

docs/developer/users.mdx
pawelpikus: Oh, shoot. It should be "we get the data...". Nice catch, thanks.

PR #552, opened 2022-08-31
Unify feature versions

peelar:


docs/dashboard/catalog/gift-cards.md
```diff
@@ -2,7 +2,7 @@
 title: Gift Cards
 ---
 
-:::note
+:::info
 Gift cards are available in Saleor 3.1.
```
patrys, line 6: Let's change this text to say "this feature was introduced..." as it does in other places. The current wording may be mistaken for "this was only available in one version."

CossackDex (approved):

PR #600, opened 2022-11-24
Add docs for metafield and metafields

patrys:
Those were introduced in Saleor 3.3 but were never documented.

### DIFF
--- a/docs/developer/metadata.mdx
+++ b/docs/developer/metadata.mdx
@@ -124,7 +124,7 @@ Response:
 
 ## Selecting fields to return
 
-By default, `metafields` (`privateMetafields`) will return all metadata that the object has assigned. You can use the optional `keys` argument to select the fields you want to return. We highly recommend that you only fetch the fields your application understands as it limits the amount of information being needlessly transferred.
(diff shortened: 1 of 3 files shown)


docs/developer/metadata.mdx
```diff
@@ -108,6 +122,44 @@ Response:
 }
 ```
 
+## Selecting fields to return
+
+By default, `metafields` (`privateMetafields`) will return all metadata that the object has assigned. You can use the optional `keys` argument to select the fields you want to return. We highly recommend that you only fetch the fields your application understands as it limits the amount of information being needlessly transferred.
```
peelar, line 127:
```suggestion
By default, `metafields` (`privateMetafields`) will return all metadata assigned to an object. You can use the optional `keys` argument to select the fields you want to return. We highly recommend that you only fetch the fields your application understands as it limits the amount of information being needlessly transferred.
```

@patrys What do you mean by "fields your application understands"?

peelar (approved):

docs/developer/metadata.mdx
lkostrowski: Maybe let's change to `uses` or `needs`? (its also not only data but interface segregation principle, not relying on fields that are not needed)

PR #604, opened 2022-11-28
Fix subject of API

ags-slc:


peelar (approved):
Thank you for your contribution 🙌

PR #614, opened 2022-12-12
remove docs 2.x

peelar:


docusaurus.config.js
```diff
@@ -47,7 +47,7 @@ module.exports = {
          */
         createRedirects: function (existingPath) {
           if (existingPath.startsWith("/docs/2.11")) {
-            return existingPath.replace("/2.11", "");
+            return existingPath.replace("/2.11", "/3.x");
```
patrys, line 50: This would redirect all 3.x docs to 2.11, it’s safe to remove this conditional branch entirely as it redirected unversioned docs to 2.11 when we enabled explicit versioning.

patrys (approved):

PR #616, opened 2022-12-13
Restructure overview and setup

patrys:
This flattens the structure and introduces index pages. The first two categories are automatically generated from the folder structure.

API reference now has a separate sidebar to make navigation easier.

This is step one of many. If this gets merged, I'll start to rebuild the other categories.


peelar:
I am super pumped about it, thanks a lot @patrys 👏. 

Two things, though, regarding this card:

image

1. The ones on the docs index page (above) are a bit different than the rest:
image

The space between the title and description is bigger in the first one. I think the latter looks better. Shouldn't we unify it?

2. The index page cards don't have an icon. Right now, [they do](https://docs.saleor.io/docs/3.x/). Is that intentional?

peelar (approved):

CossackDex (approved):

PR #642, opened 2023-01-27
Discourage forking

timuric:


lkostrowski (approved):

docs/developer/extending/overview.mdx
```diff
@@ -2,13 +2,22 @@
...
+:::warning
+Forking Saleor is not recommended for the following reasons:
+
+- Non trivial upgrades due to merge conflicts, incompatible business logic and data migrations.
+- Limited or non existing community support.
+- Maintenance and security fixes must be applied manually.
+- You would have to fork and maintain dashboard which is another large codebase.
+:::
+
 It may seem that the easiest way to customize Saleor is to modify its source code directly. While possible, it leads to all sorts of problems, including the inability to upgrade Saleor easily if a security fix is required. To keep your changes separate from the Core, Saleor offers two mechanisms to aid with customization:
```
peelar, line 14:
```
:::warning
Forking Saleor is not recommended for the following reasons:

- Non trivial upgrades due to merge conflicts, incompatible business logic and data migrations.
- Limited or non existing community support.
- Maintenance and security fixes must be applied manually.
- You would have to fork and maintain dashboard which is another large codebase.
:::

It may seem that the easiest way to customize Saleor is to modify its source code directly. While possible, it leads to all sorts of problems, including the inability to upgrade Saleor easily if a security fix is required. To keep your changes separate from the Core, Saleor offers two mechanisms to aid with customization:
```

To me, those two say almost the same thing. The first one (although more detailed) does it in points, the latter does it in a paragraph. Maybe we can supplement the paragraph with the points made in the `warning` section or remove the paragraph and just keep the `warning`?

timuric: removed that sentence

PR #646, opened 2023-02-01
docs: 🔐 update authentication guide

mmiszy:
Fixes https://github.com/saleor/react-storefront/issues/716

### DIFF
--- a/docs/api-usage/authentication.mdx
+++ b/docs/api-usage/authentication.mdx
@@ -97,7 +97,9 @@ A decoded token has the following structure:
 }
 ```
 
-Note: `csrf_token` field is deprecated and is not used in current authentication flow.
+:::info
+`csrf_token` field is deprecated and is not used in current authentication flow.
+:::
 
 ### Fetching information about currently authenticated user
 
@@ -114,7 +116,9 @@ You can use the [`me`](api-reference/queries/me.mdx) query to retrieve informati
 
 ### Verifying tokens
 
-Note: when implementing a storefront, you should probably use the [`me`](#fetching-information-about-currently-authenticated-user) query instead.
+:::tip
+When implementing a storefront, you should probably use the [`me`](#fetching-information-about-currently-authenticated-user) query instead.
+:::
 
 There are two ways to verify a token: by using the JWKS file or via a GraphQL mutation.
 
@@ -142,8 +146,6 @@ mutation {
 
 Refresh tokens have a longer lifetime than access tokens. You can generate a new access token by using the `tokenRefresh` mutation. It takes a refresh token as input and returns a new access token.
 
-Note: Keep in mind that you should not add the `Authorization` header with an expired access token to this request. It'll result in an error.
-
 ```graphql
 mutation {
   tokenRefresh(refreshToken: "<refresh-token>") {
@@ -151,6 +153,10 @@ mutation {
   }
 }
 ```
+
+:::caution
+Keep in mind that you should **not** add the `Authorization` header with an expired access token to this request. It'll result in an error.
+:::
 
 In the future, `tokenRefresh` mutation will also generate a new refresh token, and refresh tokens will be one-time use only.
 
--- a/docs/api-usage/authentication.mdx
+++ b/docs/api-usage/authentication.mdx
@@ -76,11 +76,55 @@ A decoded token has the following structure:
 }
 ```
 
+### Refresh token structure
+
+Saleor's refresh tokens are [JSON Web Tokens](https://jwt.io) signed with RS256.
+
+A decoded token has the following structure:
+
+```json
+{
+  "iat": 1671039993,
+  "owner": "saleor",
+  "iss": "http://demo.saleor.io/graphql/",
+  "exp": 1673631993,
+  "token": "K3IJEFFZgshc",
+  "email": "admin@example.com",
+  "type": "refresh",
+  "user_id": "VXNlcjo1OQ==",
+  "is_staff": true,
+  "csrf_token": "cAgshsjNr9AnsMKsAkpq4Q3fb4qBBPQT5mIyDMxShQEyO1fVpwCVzVyNQAJAZfKi"
+}
+```
+
+Note: `csrf_token` field is deprecated and is not used in current authentication flow.
+
+### Fetching information about currently authenticated user
+
+You can use the [`me`](api-reference/queries/me.mdx) query to retrieve information about the currently authenticated user:
+
+```graphql
+{
+  me {
+    id
+    email
+  }
+}
+```
+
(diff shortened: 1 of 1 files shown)


docs/api-usage/authentication.mdx
```diff
@@ -76,25 +76,34 @@ A decoded token has the following structure:
...
+  "csrf_token": "cAgshsjNr9AnsMKsAkpq4Q3fb4qBBPQT5mIyDMxShQEyO1fVpwCVzVyNQAJAZfKi"
 }
 ```
 
-You can also use the [`me`](api-reference/queries/me.mdx) query to retrieve information about the currently authenticated user:
+:::info
+`csrf_token` field is deprecated and is not used in current authentication flow.
+:::
+
+### Fetching information about currently authenticated user
```
peelar, line 104:
```suggestion
### Fetching information about the current user
```

IMO shorter = easier to parse in the navigation

docs/api-usage/authentication.mdx
```diff
@@ -76,25 +76,34 @@ A decoded token has the following structure:
...
+
+You can use the [`me`](api-reference/queries/me.mdx) query to retrieve information about the currently authenticated user:
```
peelar, line 106:
```
You can use the [`me`](api-reference/queries/me.mdx) query to retrieve information about the currently authenticated user:
```

I know it's not the object of this PR, but could you please change the query snippet below from:
```graphql
{
  me {
    id
    email
  }
}
```

to:

```graphql
query {
  me {
    id
    email
  }
}
```

I think since we always use the `mutation` keyword in our mutation snippets, we should be consistent with queries as well. And it reads better, IMO.

docs/api-usage/authentication.mdx
```diff
@@ -105,9 +114,53 @@ You can also use the [`me`](api-reference/queries/me.mdx) query to retrieve info
 }
 ```
 
+### Verifying tokens
+
+:::tip
+When implementing a storefront, you should probably use the [`me`](#fetching-information-about-currently-authenticated-user) query instead.
```
peelar, line 120: The sentence says: "you should probably use the `me` query instead", but it's unclear to me what we should use it for.

We could use some context for why these two ("verifying tokens" and "me query") are related to each other. 

I think the structure of the article is the source of my confusion. From such a flat structure, it's difficult to tell relationships between sections:

image

Maybe those two paragraphs should be nested under one heading?

If you have any idea for how we can better structure this whole document, feel free to suggest it, and I can implement it in the next PR.

docs/api-usage/authentication.mdx
```diff
@@ -105,9 +114,53 @@ You can also use the [`me`](api-reference/queries/me.mdx) query to retrieve info
...
+:::
+
+There are two ways to verify a token: by using the JWKS file or via a GraphQL mutation.
+
+#### Verifying tokens with a JWKS file
+
+You can verify tokens in your code by checking that it's signed with a trusted key. You can find the necessary JSON Web Key Set (JWKS) by visiting `https://<your-saleor-domain>/.well-known/jwks.json`.
+
+Example implemented in TypeScript with use of the `jose` library:
```
peelar, line 129:
```suggestion
Here is an example implemented in TypeScript with the use of the `jose` library:
```

docs/api-usage/authentication.mdx
```diff
@@ -117,6 +170,12 @@ mutation {
 }
 ```
 
+:::caution
+Keep in mind that you should **not** add the `Authorization` header with an expired access token to this request. It'll result in an error.
```
peelar, line 174:
```suggestion
Keep in mind that you should **not** add the `Authorization` header with an expired access token to this request. It will result in an error.
```

docs/api-usage/authentication.mdx
```diff
@@ -133,6 +192,46 @@ mutation {
 }
 ```
 
+## Typical authentication flow for a storefront
+
+In the most common scenario, you'll use the `tokenCreate` mutation with user's email and password when they try logging in inside a storefront.
```
peelar, line 197:
```suggestion
In the most common scenario, you'll use the `tokenCreate` mutation with the user's email and password when they try logging in inside a storefront.
```

docs/api-usage/authentication.mdx
```diff
@@ -133,6 +192,46 @@ mutation {
...
+
+You'll receive an access token and a refresh token. Depending on your use-case, you might consider **storing refresh token** in `localStorage` so that the token is persisted between page reloads and browser sessions.
```
peelar, line 199:
```suggestion
The mutation will return an access token and a refresh token. Depending on your use case, you might consider **storing refresh token** in `localStorage` so that the token is persisted between page reloads and browser sessions.
```

docs/api-usage/authentication.mdx
```diff
@@ -133,6 +192,46 @@ mutation {
...
+
+You should make sure to [refresh your access token](#refreshing-the-access-token) (generate a new one using refresh token) whenever it's close to its expiration date or expired.
+
+If you make a request with an expired access token, you'll receive `ExpiredSignatureError` ("Signature has expired"). You can implement a flow where your app would automatically try to refresh the access token in such cases, and then repeat the same request with the newly generated access token.
```
peelar, line 203:
```suggestion
If you make a request with an expired access token, you will receive an `ExpiredSignatureError` ("Signature has expired"). You can implement a flow where your app automatically tries to refresh the access token in such cases, and then repeat the same request with the newly generated access token.
```

docs/api-usage/authentication.mdx
```diff
@@ -133,6 +192,46 @@ mutation {
...
+
+When refreshing access token fails, you should remove the refresh token from storage and ask user to log in again.
```
peelar, line 205:
```suggestion
When refreshing the access token fails, you should remove the refresh token from storage and ask the user to log in again.
```

peelar (changes requested):
Great effort 👏. I am certain this will be a valuable read for our users. I especially like your "Typical authentication flow for a storefront" paragraph, as it breaks down this wall of theoretical content and gives the whole article practical value. 

I am aware this is not the scope of your PR, but I would like to hear your opinion.

There are plenty of valuable insights in this article, but overall I feel a bit overwhelmed by it. To me, it misses some general structure, as it looks like we just wrote down everything we know about Saleor auth (almost like in points).

If this is not just my impression, we can iterate over it in the next PRs.

docs/api-usage/authentication.mdx
mmiszy: I specifically used the term `currently authenticated user` because it was already used in several other places on this page.

peelar (approved):

PR #650, opened 2023-02-05
chore: suggest alternative for non-tty environment

gempain:
Suggest an alternative method of creating the superuser account when Saleor was deployed from a non-tty environment (e.g. Github Action).

### DIFF
--- a/docs/setup/docker-compose.mdx
+++ b/docs/setup/docker-compose.mdx
@@ -50,6 +50,13 @@ Finally, create an admin account:
 docker-compose run --rm api python3 manage.py createsuperuser
 ```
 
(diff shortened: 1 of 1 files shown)


peelar:
Hello @gempain 👋 

Big thanks for your contribution. Unfortunately, we want to keep the documentation pretty general. The proposed changes seem to be a bit too specific to your use case.

BTW. in a non-interactive environment, you can pass `--createsuperuser` in the `populatedb` command.

gempain:
> BTW. in a non-interactive environment, you can pass `--createsuperuser` in the `populatedb` command.

Unless I am mistaken, this creates a default user with a simple password and email, it seems a bit unsafe, but surely you can quickly log in and change the password.

PR #653, opened 2023-02-08
Fix missing backticks in docs/setup/configuration.mdx file

Nakaoni:
Fix missing backtick in the documentation which broke the code highlighting.


peelar:
Good catch @Nakaoni 🙇. Thanks!

peelar (approved):

PR #663, opened 2023-02-17
Add filtering page

IKarbowiak:
Add a filtering page describing what was introduced in [this PR](https://github.com/saleor/saleor/pull/11737).

### DIFF
--- a/docs/developer/filtering.mdx
+++ b/docs/developer/filtering.mdx
@@ -4,7 +4,7 @@ title: Filtering
 
 :::info
 
-The new filtering approach was introduced in **Saleor 3.11**. For now, it's only
+The "where" filtering approach was introduced in **Saleor 3.11**. For now, it's only
 available on the `Attribute` type.
 
 :::
@@ -18,25 +18,25 @@ and subject to change but is available to experiment and provide feedback.
 
 ## Introduction
 
-New filtering allows specifying more complex conditions with the use of `AND` and `OR` operators,
+The "where" filtering allows specifying more complex conditions with the use of `AND` and `OR` operators,
 combined with operations like equal (`eq`), one of, and range. It's available under the `where` option,
-for now, only on attributes query.
+for now, only on the [attributes](api-reference/queries/attributes.mdx) query.
 
 ### General approach
 
 - Currently only `AND` and `OR` operators are available.
   (See the example for [AND](#filtering-with-the-and-operator) and [OR](#filtering-with-the-or-operator) operator usage).
-- It's allowed to filter by flat fields like it was before, as well as specifying operators,
-  but this approach cannot be mixed. ([See the example](#mixing-flat-field-filter-and-operator-and-or)).
-- Only one operator can be specified on each level. ([See the example](#mixing-operators-on-the-same-level)).
-- In the new approach we decided to add a filter option for each field of the given type.
-  Because of that, the `search` filter is moved to the top level. ([See the example](##search-filtering)).
-- The filter input type matches the property name in the API type.
+- You can filter by flat fields and specify operators, but you cannot mix these two approaches.
+  ([See the example](#mixing-flat-field-filter-and-operator-and-or)).
+- At each level, you can specify only one operator. ([See the example](#mixing-operators-on-the-same-level)).
+- In the new "where" approach, each field of a given type has a corresponding filter option.
+  As a result, the "search" filter is now located at the top level. ([See the example](##search-filtering)).
+- The filter input type corresponds to the property name in the API type.
 - The string, enum, date, and all numeric type fields have specific input types
   with available operations options. Currently there is option to filter by specific
   value ([example](#filtering-by-a-specific-value)) and list of values ([example](#filtering-by-a-list-of-values)).
-- Only one operation is allowed to specify. Only `eq` or `oneOf`, providing both will raise an error. ([See the example](#mixing-different-operations)).
-- It's not allowed to provide `filter` and `where` options at the same time. ([See the example](#mixing-filter-and-where-options)).
+- You can only specify one operation. Using eq or oneOf is acceptable, but using both will result in an error. ([See the example](#mixing-different-operations)).
+- You cannot provide both `filter` and `where` options at the same time. ([See the example](#mixing-filter-and-where-options)).
 
 ## Examples
 
@@ -110,7 +110,7 @@ query {
 
 ### Filtering with the `OR` operator
 
-The `OR` operator allows defining a list of conditions that at least one must be fulfilled
+The `OR` operator allows you to define a list of conditions, of which at least one must be fulfilled.
 
 > **_Example:_** _Return attributes that have `Author` name, or input type `DATE` or `DATE_TIME`._
 
@@ -168,7 +168,7 @@ query {
 
 ### Search filtering
 
-There is still a possibility to search objects. The search is available on the root level.
+To use the full-text search capabilities, use the root-level `search` argument.
 
 > **_Example:_** _Search attributes by `size` value_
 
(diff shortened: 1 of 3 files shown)


fowczarek (approved):

docs/developer/filtering.mdx
```diff
@@ -0,0 +1,311 @@
...
+
+New filtering allows specifying more complex conditions with the use of `AND` and `OR` operators,
+combined with operations like equal (`eq`), one of, and range. It's available under the `where` option,
+for now, only on attributes query.
+
+### General approach
+
+- Currently only `AND` and `OR` operators are available.
+  (See the example for [AND](#filtering-with-the-and-operator) and [OR](#filtering-with-the-or-operator) operator usage).
+- It's allowed to filter by flat fields like it was before, as well as specifying operators,
```
maarcingebala, line 29: I think we shouldn't write as if this was a changelog, but as it is now.

```suggestion
- It's allowed to filter by flat fields as well as specifying operators,
```

docs/developer/filtering.mdx
```diff
@@ -0,0 +1,311 @@
+---
+title: Filtering
+---
+
+:::info
+
+The new filtering approach was introduced in **Saleor 3.11**. For now, it's only
```
maarcingebala, line 7: Maybe let's call it "where filtering", what do you think?

```suggestion
The "where" filtering approach was introduced in **Saleor 3.11**. For now, it's only
```

docs/developer/filtering.mdx
```diff
@@ -0,0 +1,311 @@
...
+This feature is currently in **Feature Preview**. This part of Saleor is not complete
+and subject to change but is available to experiment and provide feedback.
+
+:::
+
+## Introduction
+
+New filtering allows specifying more complex conditions with the use of `AND` and `OR` operators,
+combined with operations like equal (`eq`), one of, and range. It's available under the `where` option,
+for now, only on attributes query.
```
maarcingebala, line 23:
``` suggestion
The "where" filtering allows specifying complex conditions with the use of `AND` and `OR` operators,
combined with operations like equal (`eq`), one of, and range. It's available under the `where` option,
for now, only on the [attributes](api-reference/queries/attributes.mdx) query.
```

docs/developer/filtering.mdx
```diff
@@ -0,0 +1,311 @@
...
+
+### General approach
+
+- Currently only `AND` and `OR` operators are available.
+  (See the example for [AND](#filtering-with-the-and-operator) and [OR](#filtering-with-the-or-operator) operator usage).
+- It's allowed to filter by flat fields like it was before, as well as specifying operators,
+  but this approach cannot be mixed. ([See the example](#mixing-flat-field-filter-and-operator-and-or)).
+- Only one operator can be specified on each level. ([See the example](#mixing-operators-on-the-same-level)).
+- In the new approach we decided to add a filter option for each field of the given type.
```
maarcingebala, line 32: I would rephrase this to not sound like a changelog, but a current state.

docs/developer/filtering.mdx
```diff
@@ -0,0 +1,311 @@
+---
+title: Filtering
+---
+
+:::info
+
+The "where" filtering approach was introduced in **Saleor 3.11**. For now, it's only
+available on the `Attribute` type.
```
peelar, line 8:
```suggestion
The `where` filtering approach was introduced in **Saleor 3.11**. For now, it's only
available on the `Attribute` type.
```

Also, regarding that:
>  For now, it's only available on the `Attribute` type.

Isn't it the [`attributes`](https://docs.saleor.io/docs/3.x/api-reference/queries/attributes) query?

docs/developer/filtering.mdx
```diff
@@ -0,0 +1,311 @@
...
+:::caution
+
+This feature is currently in **Feature Preview**. This part of Saleor is not complete
+and subject to change but is available to experiment and provide feedback.
+
+:::
+
+## Introduction
+
+The "where" filtering allows specifying more complex conditions with the use of `AND` and `OR` operators,
```
peelar, line 21:
```suggestion
The `where` filtering allows specifying more complex conditions with the use of `AND` and `OR` operators,
```

docs/developer/filtering.mdx
```diff
@@ -0,0 +1,311 @@
...
+for now, only on the [attributes](api-reference/queries/attributes.mdx) query.
+
+### General approach
+
+- Currently only `AND` and `OR` operators are available.
+  (See the example for [AND](#filtering-with-the-and-operator) and [OR](#filtering-with-the-or-operator) operator usage).
+- You can filter by flat fields and specify operators, but you cannot mix these two approaches.
+  ([See the example](#mixing-flat-field-filter-and-operator-and-or)).
+- At each level, you can specify only one operator. ([See the example](#mixing-operators-on-the-same-level)).
+- In the new "where" approach, each field of a given type has a corresponding filter option.
```
peelar, line 32:
```suggestion
- In the new `where` approach, each field of a given type has a corresponding filter option.
```

docs/developer/filtering.mdx
```diff
@@ -0,0 +1,311 @@
...
+  As a result, the "search" filter is now located at the top level. ([See the example](##search-filtering)).
```
peelar, line 33:
```suggestion
  As a result, the `search` filter is now located at the top level. ([See the example](##search-filtering)).
```

docs/developer/filtering.mdx
```diff
@@ -0,0 +1,311 @@
...
+- The filter input type corresponds to the property name in the API type.
+- The string, enum, date, and all numeric type fields have specific input types
+  with available operations options. Currently there is option to filter by specific
+  value ([example](#filtering-by-a-specific-value)) and list of values ([example](#filtering-by-a-list-of-values)).
```
peelar, line 37:
```suggestion
  with available operations options. Currently, there is an option to filter by specific
  value ([example](#filtering-by-a-specific-value)) and list of values ([example](#filtering-by-a-list-of-values)).
```

docs/developer/filtering.mdx
```diff
@@ -0,0 +1,311 @@
...
+- You can only specify one operation. Using eq or oneOf is acceptable, but using both will result in an error. ([See the example](#mixing-different-operations)).
```
peelar, line 38:
```suggestion
- You can only specify one operation. Using `eq` or `oneOf` is acceptable, but using both will result in an error. ([See the example](#mixing-different-operations)).
```

docs/developer/filtering.mdx
```diff
@@ -0,0 +1,311 @@
...
+        inputType
+      }
+    }
+  }
+}
+```
+
+### Filtering with the `AND` operator
+
+The `AND` operator allows defining a list of conditions that must be fulfilled.
```
peelar, line 90:
```suggestion
The `AND` operator allows for defining a list of conditions that must be met.
```

docs/developer/filtering.mdx
```diff
@@ -0,0 +1,311 @@
...
+      edges {
+          node {
+             id
+          }
+      }
+   }
+}
+```
+
+**INSTEAD:** Join both conditions into one AND filter input
```
peelar, line 210:
```suggestion
**INSTEAD:** Join both conditions into one `AND` filter input
```

peelar (changes requested):

docs/developer/filtering.mdx
IKarbowiak: Like I mean here that it's done for `Attributes` type with reference to the whole idea of adding the filtering input for each `Attribute` field, and adding a new filter on `attributes` query.

peelar (approved):

PR #671, opened 2023-02-27
Add discounts page

IKarbowiak:
Add the `discounts` page explaining the basics of discounts.

The page should be extended later with:
- explanation of how to create a `sale` and `voucher`
- the `shipping` voucher example

### DIFF
--- a/docs/developer/discounts.mdx
+++ b/docs/developer/discounts.mdx
@@ -75,9 +75,9 @@ mutation {
 ```
 
 Please check the line prices in the following response.
-The line.totalPrice is 8.1, and the line.undiscountedTotalPrice is 9.0.
+The line.totalPrice is **8.1**, and the `line.undiscountedTotalPrice` is **9.0**.
 
-```json {16-37}
+```json {7,16-37}
 {
   "data": {
     "checkoutLinesAdd": {
@@ -207,7 +207,7 @@ mutation {
 
 Here is the response:
 
-```json
+```json {11,27-37,52-62}
 {
   "data": {
     "checkoutAddPromoCode": {
@@ -310,7 +310,7 @@ Refer to the response from running the same mutation as before.
 The `checkout.discount` is **4.0**. The cheapest line's totalPrice is **0.0**,
 and the undiscountedTotalPrice is **4.0**.
 
-```json {27-37}
+```json {11,27-37}
 {
 	"data": {
 		"checkoutAddPromoCode": {
@@ -397,7 +397,7 @@ In the following example, a **10%** voucher for a specific product is applied du
 The discount applies to the first two lines, one for **$45** and the other for **$20**, but not to the third line for **$1.99**.
 The response of running the `checkoutAddPromoCode` mutation with the voucher code for this discount is shown below:
 
-```json
+```json {11,27-37,52-62,77-87}
 {
   "data": {
     "checkoutAddPromoCode": {
@@ -514,7 +514,7 @@ If the voucher has the `applyOncePerOrder` flag turned on, the discount will onl
 the single cheapest product that is eligible for the discount. In the scenario described,
 the discount would only be applied to the product with a price of **$20**, and would be visible on only one line of the order.
 
-```json
+```json {11,27-37,52-62,77-87}
 {
   "data": {
     "checkoutAddPromoCode": {
@@ -633,7 +633,7 @@ A fixed amount discount of $5 is being applied to the entire order.
 
 Here is the checkout data before applying the voucher code (only the sale is included in the price).
 
-```json
+```json {11,34-44,56-66}
 {
   "data": {
     "checkout": {
@@ -723,7 +723,7 @@ is **$0.0**.
 
 Below is the checkout data after applying the entire order voucher.
 
-```json
+```json {11,27-36,52-62}
 {
   "data": {
     "checkoutAddPromoCode": {
@@ -801,14 +801,7 @@ Below is the checkout data after applying the entire order voucher.
         ]
       }
     }
-  },
-  "extensions": {
-    "cost": {
-      "requestedQueryCost": 1,
-      "maximumAvailable": 50000
-    }
   }
-}
 ```
 
 As we can see, the total discount is **$5.0**. For the first line,
--- a/docs/developer/discounts.mdx
+++ b/docs/developer/discounts.mdx
@@ -1,0 +1,175 @@
+---
+title: Discounts
+sidebar_label: Discounts
+---
+
+## Introduction
+
+There are two kinds of discounts in Saleor: `Sales` and `Vouchers`.
+Discounts allow for the reduction of prices for selected variants, products, collections,
+or categories by a given percentage or fixed value.
+The sale discount is automatically applied to all included products, without requiring
+any additional actions from the user. In contrast, vouchers require the user to provide
+a code during the checkout process. Both the sale and voucher discounts can be used together.
+
+### Sale
+
+As mentioned earlier, sales are automatically applied when a product is added to the checkout.
+For instance, if a product is priced at **$9** and has a **10%** discount, it will be available for **$8.1**.
+The discount will be visible on line prices. To calculate the total applied discount,
+subtract the `line.undiscountedTotalPrice` from the `line.totalPrice`.
+To calculate the discount on one unit, subtract the `line.undiscountedUnitPrice` from the `line.unitPrice`.
+
+:::info
+The sale discount is not visible on `checkout.discount.amount`.
+:::
+
+Consider the following example: after adding a product on sale to the checkout,
+the response will include the discounted price of the product.
+
+```graphql
+mutation {
+  checkoutLinesAdd(
+    checkoutId: "Q2hlY2tvdXQ6NzMwMzkwMmItZGRhOC00MzU3LTk1YTAtNjRiODNiNTllMmUy"
+    lines: { quantity: 1, variantId: "UHJvZHVjdFZhcmlhbnQ6Mzcx" }
+  ) {
+    errors {
+      message
+    }
+    checkout {
+      discount {
+        amount
+      }
+      token
+      lines {
+        quantity
+        variant {
+          name
+        }
+        totalPrice {
+          gross {
+            amount
+          }
+          net {
+            amount
+          }
+        }
+        undiscountedTotalPrice {
+          amount
+        }
+        unitPrice {
+          gross {
+            amount
+          }
+          net {
+            amount
+          }
+        }
+        undiscountedUnitPrice {
+          amount
+        }
+      }
+    }
+  }
+}
+```
+
+Please check the line prices in the following response.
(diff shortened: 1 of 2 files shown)


docs/developer/discounts.mdx
```diff
@@ -0,0 +1,815 @@
...
+---
+
+## Introduction
+
+There are two kinds of discounts in Saleor: `Sales` and `Vouchers`.
+Discounts allow for the reduction of prices for selected variants, products, collections,
+or categories by a given percentage or fixed value.
+The sale discount is automatically applied to all included products, without requiring
+any additional actions from the user. In contrast, vouchers require the user to provide
+a code during the checkout process. Both the sale and voucher discounts can be used together.
```
peelar, line 13:
```suggestion
Discounts allow for the reduction of prices for selected variants, products, collections,
or categories by a given percentage or a fixed value.
There are two kinds of discounts in Saleor: `Sales` and `Vouchers`.
The **sale** discount is automatically applied to all included products, without requiring
any additional actions from the user. In contrast, **vouchers** require the user to provide
a code during the checkout process. Both the sale and voucher discounts can be used together.
```

>all included products
We are missing where these products are included.

docs/developer/discounts.mdx
```diff
@@ -0,0 +1,815 @@
...
+
+### Sale
+
+As mentioned earlier, sales are automatically applied when a product is added to the checkout.
+For instance, if a product is priced at **$9** and has a **10%** discount, it will be available for **$8.1**.
+The discount will be visible on the line prices. To calculate the total applied discount,
+subtract the `line.undiscountedTotalPrice` from the `line.totalPrice`.
+To calculate the discount on one unit, subtract the `line.undiscountedUnitPrice` from the `line.unitPrice`.
```
peelar, line 21:
```suggestion
Sales are automatically applied when a product is added to the checkout.
For instance, if a product is priced at **$9** and has a **10%** discount, it will be available for **$8.1**.
The discount will be visible on the line prices. To calculate the total applied discount,
subtract the `line.undiscountedTotalPrice` from the `line.totalPrice`.
To calculate the discount on one unit, subtract the `line.undiscountedUnitPrice` from the `line.unitPrice`.
```

docs/developer/discounts.mdx
```diff
@@ -0,0 +1,815 @@
...
+
+:::info
+The sale discount is not visible on `checkout.discount.amount`.
+:::
+
+Consider the following example: after adding a product on sale to the checkout,
+the response will include the discounted price of the product.
```
peelar, line 28:
```suggestion
In the example below, we add an on-sale product to the checkout. The response of the `checkoutLinesAdd` mutation will include the discounted product price.
```

docs/developer/discounts.mdx
```diff
@@ -0,0 +1,815 @@
...
+          amount
+        }
+      }
+    }
+  }
+}
+```
+
+Please check the line prices in the following response.
+The line.totalPrice is **8.1**, and the `line.undiscountedTotalPrice` is **9.0**.
```
peelar, line 78:
```suggestion
The `line.totalPrice` is **8.1**, and the `line.undiscountedTotalPrice` is **9.0**.
```

docs/developer/discounts.mdx
```diff
@@ -0,0 +1,815 @@
...
+        undiscountedUnitPrice {
+          amount
+        }
+      }
+    }
+  }
+}
+```
+
+Please check the line prices in the following response.
```
peelar, line 77:
```suggestion
Please note the line prices in the following response.
```

docs/developer/discounts.mdx
```diff
@@ -0,0 +1,815 @@
...
+
+There are three types of vouchers:
+
+- Fixed amount: reduces the price by a specified value.
+- Percentage: reduces the price by a specified percentage value.
+- Shipping: reduces the shipping price.
+
+The voucher can be applied to all checkout products, or only to specified ones.
+
+There is also an option to apply the discount only to the cheapest eligible product.
```
peelar, line 136:
```
There is also an option to apply the discount only to the cheapest eligible product.
```

What does "eligible" mean here?

docs/developer/discounts.mdx
```diff
@@ -0,0 +1,815 @@
...
+```
+
+The discount is visible on both the `checkout.discount` field and the prices in `checkout.lines`.
+In the first line, the `totalPrice` is **$0.41** less than the `undiscountedTotalPrice`,
+while the difference in the second line is **$4.59**.
+The `checkout.discount` is the sum of the differences between `totalPrice` and `undiscountedTotalPrice` of the lines.
+
+:::info
+If an entire order voucher with a fixed amount is applied during checkout,
+and the order contains multiple lines, the discount will be distributed evenly in proportion to the total price of each line.
```
peelar, line 299:
```suggestion
If the user applied a fixed-amount order voucher during checkout, and the order contains multiple lines, the discount will be distributed evenly in proportion to the total price of each line.
```

docs/developer/discounts.mdx
```diff
@@ -0,0 +1,815 @@
...
+:::
+
+#### Applying the entire order voucher that is applied once per order
```
peelar, line 302:
```suggestion
#### Applying the entire order voucher once per order
```

Not sure if my suggestion is valid, but we should try to rephrase that.

docs/developer/discounts.mdx
```diff
@@ -0,0 +1,815 @@
...
+
+If a voucher with the `applyOncePerOrder` flag set to True is used in a similar scenario,
+the discount will only apply to the cheapest eligible product. In this checkout,
+the cheapest eligible product is priced at **$4**.
+Therefore, the discount amount will also be $4 and will only appear on one line.
```
peelar, line 307:
```suggestion
If a voucher with the `applyOncePerOrder` flag set to `True` is used in a similar scenario,
the discount will only apply to the cheapest eligible product. In this checkout,
the cheapest eligible product is priced at **$4**.
Therefore, the discount will be $4 and will only appear on one line.
```

docs/developer/discounts.mdx
```diff
@@ -0,0 +1,815 @@
...
+
+Refer to the response from running the same mutation as before.
+The `checkout.discount` is **4.0**. The cheapest line's totalPrice is **0.0**,
+and the undiscountedTotalPrice is **4.0**.
```
peelar, line 311:
```suggestion
Refer to the response from running the same mutation as before.
The `checkout.discount` is **4.0**. The cheapest line's `totalPrice` is **0.0**,
and the `undiscountedTotalPrice` is **4.0**.
```

docs/developer/discounts.mdx
```diff
@@ -0,0 +1,815 @@
...
+    }
+  }
+}
+```
+
+As we can see, the 10% discount has been applied to the first two lines.
+The total discount amount is visible in the `checkout.discount` field, which is equal to
+the sum of the differences between the `undiscountedTotalPrice` and `totalPrice` of all lines.
+
+#### Applying the specific product voucher that applies once per order
```
peelar, line 511: same as above, too much "applying"

docs/developer/discounts.mdx
```diff
@@ -0,0 +1,815 @@
...
+
+If the voucher has the `applyOncePerOrder` flag turned on, the discount will only be applied to
+the single cheapest product that is eligible for the discount. In the scenario described,
+the discount would only be applied to the product with a price of **$20**, and would be visible on only one line of the order.
```
peelar, line 515:
```suggestion
If the voucher has the `applyOncePerOrder` flag set to `True`, the discount will only be applied to
the single cheapest product eligible for the discount. In the scenario described,
the discount would only be applied to the product with a price of **$20**, and would be visible on only one line of the order.
```

docs/developer/discounts.mdx
```diff
@@ -0,0 +1,815 @@
...
+        ]
+      }
+    }
+  }
+}
+```
+
+Here, we can see that the discount was only applied to the cheapest line included in the voucher discount.
+The total `checkout.discount` is only **$2.0** in this case. The difference between
+`totalPrice` and `undiscountedTotalPrice` is only visible on the second line.
```
peelar, line 626:
```suggestion
Here, we can see that the discount was only applied to the cheapest line included in the voucher discount.
The total `checkout.discount` is **$2.0** in this case. The difference between
`totalPrice` and `undiscountedTotalPrice` is only visible on the second line.
```

docs/developer/discounts.mdx
```diff
@@ -0,0 +1,815 @@
...
+
+### Sale and Voucher together
+
+Sales and vouchers can be combined. In this case, the voucher discount is applied to the price after the sale.
+Let's consider an example: the checkout has two items, and the second item is on sale.
+A fixed amount discount of $5 is being applied to the entire order.
```
peelar, line 632:
```suggestion
A fixed discount of **$5** is being applied to the entire order.
```

docs/developer/discounts.mdx
```diff
@@ -0,0 +1,815 @@
...
+      }
+    }
+  }
+```
+
+As we can see, the total discount is **$5.0**. For the first line,
+the entire difference between `totalPrice` and `undiscountedTotalPrice` comes from
+the voucher discount ($1.94). However, in the case of the second line, the difference
+between those values ($6.56) is the total discount applied to this line,
+which includes both sales and voucher discounts.
```
peelar, line 811:
```suggestion
As we can see, the total discount is **$5.0**. In the first line,
the entire difference between `totalPrice` and `undiscountedTotalPrice` comes from
the voucher discount ($1.94). However, in the second line, the difference
between those values ($6.56) is the total discount applied to this line,
which includes both sales and voucher discounts.
```

docs/developer/discounts.mdx
IKarbowiak: I mean the product that is included in the voucher. But maybe it's not needed here?

docs/developer/discounts.mdx
```diff
@@ -0,0 +1,815 @@
...
+  }
+```
+
+As we can see, the total discount is **$5.0**. In the first line,
+the entire difference between `totalPrice` and `undiscountedTotalPrice` comes from
+the voucher discount ($1.94). However, in the second line, the difference
+between those values ($6.56) is the total discount applied to this line,
+which includes both sales and voucher discounts.
+
+To calculate the value of the sale that was applied, we can sum up the line discounts and subtract
```
peelar, line 813:
```suggestion
To calculate the value of the applied sale, we can sum up the line discounts and subtract
```

docs/developer/discounts.mdx
```diff
@@ -0,0 +1,815 @@
...
+`checkout.discount`. In this example, the calculation would be: **($1.94 + $6.56) - $5.00 = $3.5**.
+Therefore, we end up with the same value as calculated previously from the response, before the voucher was applied.
```
peelar, line 815:
```suggestion
Therefore, we end up with the same value as before applying the voucher.
```

peelar (approved):

PR #678, opened 2023-03-10
Add `requiredSaleorVersion` and `author` fields to App manifest

przlada:
Add new fields `requiredSaleorVersion` and `author` to the App manifest description


peelar:
@przlada Do you think it makes sense to include information about the version of Saleor these fields were introduced in?

przlada:
@peelar Great question 👍 Those fields are optional and, if present, will not break compatibility with earlier Saleor versions (will be ignored). But my reasoning was not to highlight this optionality and encourage app developers to use them even thou they will start to be supported in 3.13. But I am open to discussion😄

peelar (approved):

PR #679, opened 2023-03-10
Inject examples to api-reference

2can:
This PR introduces the `Examples` section to the API reference pages. 

image

### How it works
Example section is put in the generated API reference page using [Markdown imports](https://docusaurus.io/docs/markdown-features/react#importing-markdown). The import is put into the page with the use of the `inject-examples` script. The script takes the corresponding file with examples and imports into the page.  

The `update-reference` workflow has been modified to do that.

### How to add examples
Pages with examples are stored in a separate `examples` folder. 
This folder structure should be the same as the one generated with the [GraphQL-Markdown](https://graphql-markdown.github.io/) `api-reference`. 

1. Create a new file in the `examples` folder with the same path as the API reference page you are documenting.
1. Import the file to the API reference page eg.

```
import Product from "@site/examples/docs/api-reference/queries/product.mdx";

<Product />
```

### DIFF
--- a/scripts/inject-examples.ts
+++ b/scripts/inject-examples.ts
@@ -1,0 +1,36 @@
+const fg = require("fast-glob");
+import * as fs from "node:fs/promises";
+
+function filePathToPascalCase(path: string): string {
+  const nameRe = /\/(\w+)\.mdx/;
+  const name = (path.match(nameRe) || ["", ""])[1];
+  const camelCaseStr = name.replace(/-([a-z])/g, function (g) {
+    return g[1].toUpperCase();
+  });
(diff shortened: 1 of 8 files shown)


scripts/inject-examples.ts
```diff
@@ -0,0 +1,37 @@
+const fg = require("fast-glob");
+import * as fs from "node:fs/promises";
+
+function filePathToPascalCase(path: string): string {
+  const nameRe = /\/(\w+)\.mdx/;
+  const name = (path.match(nameRe) || ["", ""])[1];
+  console.log;
```
peelar, line 7:
```suggestion
```

peelar (approved):

PR #709, opened 2023-04-04
Change required permissions text

2can:


patrys:
Most operations only require holding one of the listed permissions. This change makes it look like all are needed.

PR #716, opened 2023-04-05
Add "updating domain configuration"

przlada:
I want to add this change as it brings the "Updating domain configuration" and "ENABLE_SSL" paragraphs to "Server Configuration"

### DIFF
--- a/docs/setup/configuration.mdx
+++ b/docs/setup/configuration.mdx
@@ -134,6 +134,10 @@ Controls whether to run [Django Debug Toolbar](https://django-debug-toolbar.read
 To use the toolbar, you must also enable the `PLAYGROUND_ENABLED` setting. Then, the toolbar will be rendered when accessing the `/graphql/` URL.
 Note that using the toolbar is recommended only locally and should be disabled in production, as it affects the performance and may potentially reveal sensitive or private information.
 
+### `ENABLE_SSL`
+
+Used while generating URLs. Set to `True` if Saleor is deployed behind HTTPS. Defaults to `False`.
+
 ### `EMAIL_URL`
 
 The URL of the email gateway. Defaults to printing everything to the console.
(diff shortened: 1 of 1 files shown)


docs/setup/configuration.mdx
```diff
@@ -236,3 +240,24 @@ Saleor can automatically convert amounts between currencies. To use this feature
...
+     shopDomainUpdate(input: { domain: "api.example.com" }) {
+       shop {
+         domain {
+           host
+         }
+       }
+     }
+   }
+   ```
+   Note that Saleor instances must be restarted for the new configuration to take effect, as shop data is cached in memory.
```
patrys, line 260: I don't believe that's true.

przlada: From what I've checked, the `Site` objects are stored in memory, and whenever `Site.objects.get_current()` is called, the object is retrieved from the cache if available. However, if Saleor is operating on multiple processes/workers, this can result in `Site.objects.get_current()` returning outdated information (for example, for URL generation). I suggested this "restarting" solution, but I could be missing some crucial details.

patrys: Saleor used to disable the cache. If the cache is still in effect, please file a critical bug. There is no way to restart Saleor in a clustered cloud environment.

przlada: https://github.com/saleor/saleor/issues/12528

2can:
I've added `blocked` label as I assume it's not merge-able , please remove the label while it's ready do be merged

docs/setup/configuration.mdx
2can, line 260: @patrys @przlada is this note the only thing that blocks this PR from being merged ? Can we exclude this line and merge ?

patrys: The note is accurate, meaning we're documenting poor architecture, but that's still better than not having documentation. A proper fix will come in https://github.com/saleor/saleor/issues/12748

2can (approved):

PR #726, opened 2023-04-10
Redesign docs [Do not merge]

timuric:


2can (approved):

mmiszy (approved):

patrys:
Dark mode seems broken

image

timuric:
> Dark mode seems broken
> 
> image

@patrys deliberately left like that, we are going to drop card overview screens as they are not useful and create inconsistent navigation patterns. This PR needs structural changes before merging

patrys:
@timuric 

>deliberately left like that

Why are only index pages broken though?

zaiste (approved):

timuric:
> @timuric
> 
> > deliberately left like that
> 
> Why are only index pages broken though?

Index pages don't have swizzled footer component, which is used to hook into theme events. Global theme would be moved to search component as it is present on all screens and would be more reliable.

patrys:
@timuric I don't understand why we're using custom components with React hooks to theme the page. Docusaurus supports theming without any special hacks.

PR #738, opened 2023-04-17
Regenerate API reference

2can:
This PR:
- generates API reference from scratch in order to update categories which are used for menu items sort
- updates the API reference to the newest one
- improves `sortSidebarItems`


patrys:
@2can It removes numerous files from the "misc" category but does not add them anywhere else. Is that on purpose?

2can:
@patrys yes, it's on purpose, those files were already moved from "misc" to other categories with #733 . Somehow the `update-api-referece` script didn't do a cleanup and didn't update all `_category.yml` files used for sorting.

timuric (approved):

PR #742, opened 2023-04-19
Redesign

timuric:


2can (approved):

docs/developer/payments.mdx
```diff
@@ -669,7 +669,7 @@ The webhook will be sent only to the app that is the owner of the transaction.
 
 :::
 
-##### Asynchronously processing the action
```
patrys, line 672: This looks like an unrelated change.

docusaurus.config.js
```diff
@@ -348,8 +348,8 @@ module.exports = {
     image: "img/og-image.png",
 
     prism: {
-      theme: require("prism-react-renderer/themes/github"),
-      darkTheme: require("prism-react-renderer/themes/dracula"),
+      theme: require("prism-react-renderer/themes/oceanicNext"),
+      defaultLanguage: "graphql",
```
patrys, line 352: Why set a default language? I'd rather explicitly mark all fenced code examples.

src/theme/SearchBar/index.js
```diff
@@ -1,8 +1,23 @@
...
+  //Radix color variables require a class on the root element
+  useEffect(() => {
+    //Add classes for radix color variables
+    if (colorMode === "dark") {
+      document.documentElement.classList.add("dark-theme");
+    } else {
+      document.documentElement.classList.remove("dark-theme");
+    }
+  }, [colorMode]);
+};
```
patrys, line 18: It's a hack. Please add documentation on why it's here and how it works. I'm afraid that it will stop working one day and no one will be able to understand how theming ever worked.

patrys (changes requested):

zaiste (approved):

docs/developer/payments.mdx
timuric: It is, as h5 fives should not be used as we don't differentiate h5 and h4, and those are actually typos and should be h4

docusaurus.config.js
timuric: Unintended change, thanks!

src/theme/SearchBar/index.js
timuric: This hack was removed in later commit

patrys (approved):

PR #746, opened 2023-04-21
Add docs for mutation to create checkout from order

korycins:
Add docs that cover the implementation from the core PR: https://github.com/saleor/saleor/pull/12628
The RFC can be found here: https://github.com/saleor/saleor/issues/11352


docs/developer/checkout.mdx
```diff
@@ -1657,3 +1657,139 @@ Anonymous and unfinished checkouts are deleted after a specified period of time
...
+:::
+
+:::caution
+
+This feature is in the **Feature Preview** stage, which means that it is available for experimentation and
+feedback. However, it is still undergoing development and is subject to modifications.
+
+:::
+
+Saleor allows you to create a checkout from existing orders. This feature can be used when implementing the reordering functionality.
```
mmiszy, line 1676: What happens if the prices of the variants change? We should describe that here.

mmiszy (changes requested):

docs/developer/checkout.mdx
korycins: We don't copy the prices. As prices can be reduced by sale that doesn't exists etc. I will mention this here.

mmiszy (approved):

timuric (approved):

PR #752, opened 2023-04-27
Change the recipient email in sample webhook payloads for `Customer requested email change`

korycins:
The `recipient_email` that we use to send an email is the `email` from `new_email` field, not from `old_email`.


patrys:
Is this the expected behavior or are we documenting a bug?

korycins:
@patrys The docs had the incorrect email placed there. I also checked the email logic before migrating the email to plugins. In all cases, the email's `recipient_email` is a `"new.email@example.com1"`. I've found also a GH issue where it was proposed in this way. 
Ref: https://github.com/saleor/saleor/issues/4648

patrys:
That's not my question, though. I'm wondering if the behavior is correct.

timuric (approved):

PR #784, opened 2023-05-30
Add zero-downtime documentation

kadewu:
Documentation of zero-downtime migrations

### DIFF
--- a/docs/developer/community/contributing.mdx
+++ b/docs/developer/community/contributing.mdx
@@ -265,7 +265,7 @@ Use relative imports.
 ### Migrations
 
 Try to combine multiple migrations into one, but remember not to mix changes on the database with updating rows in migrations. In other words, operations that alter tables and use `RunPython` to run methods on existing data should be in separate files.
-From Saleor version 3.13 forward follow [zero-downtime policy](zero-downtime-migrations) when writing migrations.
+From Saleor version 3.13 forward follow [zero-downtime policy](docs/developer/community/zero-downtime-migrations.mdx) when writing migrations.
 
 ### Handling migrations between versions
 
--- a/docs/developer/community/contributing.mdx
+++ b/docs/developer/community/contributing.mdx
@@ -265,6 +265,7 @@ Use relative imports.
 ### Migrations
 
 Try to combine multiple migrations into one, but remember not to mix changes on the database with updating rows in migrations. In other words, operations that alter tables and use `RunPython` to run methods on existing data should be in separate files.
+From Saleor version 3.13 forward follow [zero-downtime policy](zero-downtime-migrations) when writing migrations.
 
 ### Handling migrations between versions
 
--- a/docs/developer/community/zero-downtime-migrations.mdx
+++ b/docs/developer/community/zero-downtime-migrations.mdx
@@ -2,6 +2,8 @@
 title: Zero downtime migrations
 ---
 
+Following page contains information about writing zero-downtime migrations. If you are interested in section about upgrading Saleor, please visit [upgrading guide](docs/setup/upgrading.mdx).
+
 ## The problem
 
 From Saleor 3.13 version forward, we provide a way to ensure we can safely migrate the database without interrupting the work of already running web workers.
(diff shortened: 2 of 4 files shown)


docs/developer/community/zero-downtime-migrations.mdx
```diff
@@ -0,0 +1,344 @@
...
+        I -->J[Convert migration task into migration]
+        K[Remove temporary field from logic and ORM]
+    end
+    subgraph ver_X+3
+        K -->M[Remove field from database]
+    end
+    ver_X-1 -->ver_X -->ver_X+1 -->ver_X+2 -->ver_X+3
+```
+
+e.g. **Change field type of `field` from `int` to `string` in Saleor version 3.14:**
```
maarcingebala, line 133: Maybe we could use more real examples here? E.g. in this case we could describe it based on the example of changing the `Order.number` field from int to string.

docs/developer/community/zero-downtime-migrations.mdx
```diff
@@ -0,0 +1,344 @@
...
+For the general upgrade guide see the [upgrading guide](../../setup/upgrading).
+Zero-downtime migrations can only be achieved by upgrading Saleor minor version by minor version, that means, to upgrade Saleor from version 3.11 to 3.13, you need first upgrade your workers
+to version 3.12, then proceed with upgrading to version 3.13. By upgrading from version 3.11 directly to 3.13 you will not follow zero-downtime policy.
+
+The exact steps needed to be done:
+
+- Upgrade your current minor version to the newest patch of Saleor of that version, remember to turn off Celery workers as stated in [upgrading guide](../../setup/upgrading).
+- Upgrade your current minor version of Saleor to the next one. Remember to follow the steps stated in [upgrading guide](../../setup/upgrading).
+
+### Troubleshooting: Migration with celery task fails
```
maarcingebala, line 322:
```suggestion
### Troubleshooting: Migration with Celery task fails
```

docs/developer/community/zero-downtime-migrations.mdx
```diff
@@ -0,0 +1,344 @@
...
+
+It's safe to rerun failed migration tasks in case of an error. To do that you can for example start a task from `Django shell`
```
maarcingebala, line 324:
```suggestion
It's safe to rerun failed migration tasks in case of an error. To do that, you can for example start a task from the Django shell:
```

docs/developer/community/zero-downtime-migrations.mdx
```diff
@@ -0,0 +1,344 @@
...
+
+```bash
+python manage.py shell
+```
+
+And inside the shell just import the task and schedule it, e.g.
```
maarcingebala, line 330:
```suggestion
And inside the shell just import the task and schedule it, e.g.:
```

fowczarek (approved):

docs/developer/community/zero-downtime-migrations.mdx
maarcingebala: I just meant some more real field names to illustrate the scenario better.

docs/developer/community/zero-downtime-migrations.mdx
```diff
@@ -0,0 +1,344 @@
...
+
+class Migration(migrations.Migration):
+    dependencies = []
+
+    operations = [
+        migrations.RunPython(migration, migrations.RunPython.noop)
+    ]
+```
+
+The following conditions must be met regarding the migration task:
```
IKarbowiak, line 250: Maybe we should also add a tip how to calculate the batch size?

IKarbowiak (approved):

zaiste:
@kadewu could we move this forward?

docs/developer/community/contributing.mdx
```diff
@@ -265,6 +265,7 @@ Use relative imports.
 ### Migrations
 
 Try to combine multiple migrations into one, but remember not to mix changes on the database with updating rows in migrations. In other words, operations that alter tables and use `RunPython` to run methods on existing data should be in separate files.
+From Saleor version 3.13 forward follow [zero-downtime policy](zero-downtime-migrations) when writing migrations.
```
patrys, line 268: I think we should link to files so the link gets validated. Relative paths are generated automatically during the build phase.

docs/developer/community/zero-downtime-migrations.mdx
```diff
@@ -0,0 +1,344 @@
...
+From Saleor 3.13 version forward, we provide a way to ensure we can safely migrate the database without interrupting the work of already running web workers.
+
+## Solution
+
+To achieve the goal we need to ensure that:
+
+- Old code is compatible with new database schema or new database schema is compatible with old code.
+- Migrations don't lock tables, rows etc. for no longer than one second.
+
+## Writing migrations
```
patrys, line 16: Who is the intended audience of this document?

If the audience is Saleor users, this section does not belong here as we don't recommend that anyone forks Saleor (they would be unable to upgrade).

If the audience is developers, the section on running migrations does not belong here, as it should be part of self-hosting.

kadewu: The idea was actually to provide documentation/guide for developers and users as well. This way we can easily provide link in PR's. As well anyone interested of topic can check how migrations works and how to execute them.

I was also thinking of maybe splitting this document, but at the end decided that with one document is easier to find all relevant information.

Maybe I can just provide more links to this document in proper sections?

PR #787, opened 2023-05-31
customRanking for Algolia searchParameters

2can:


mmiszy:
Thanks, I'll add more of these if I spot anything.

PR #817, opened 2023-07-04
apps docs improvement

peelar:
## Context
We received feedback from @aniav that it's not clear enough that JavaScript isn't the only option to create Saleor apps.

## Features
- I split up the "Quickstart" article into sections "JavaScript" and "Other languages", while making it clear there is an existing app ecosystem in JS.
- Added a paragraph about the app solutions in Python, linking to Mirumee's [`saleor-app-framework-python`](https://github.com/mirumee/saleor-app-framework-python).


aniav (approved):
👌

lkostrowski (approved):

PR #832, opened 2023-07-18
Update list of supported metadata objects

maarcingebala:
We can link to the ObjectWithMetadata type for the list of all supported items, which is easier to maintain than updating the list manually.


lkostrowski:
I think we can make each badge `display: block` to make them look like a sorted list - will be more readable

Zrzut ekranu 2023-07-18 o 17 31 03

Apart of that, good to have auto-generated docs 👍

lkostrowski (approved):

PR #837, opened 2023-07-20
Maintenance: Updating dependencies

2can:


patrys:
Does the bump change how the API reference is generated?

zaiste (approved):

PR #843, opened 2023-07-25
Add details about stock tracking

timuric:


korycins (approved):

docs/developer/checkout/finalizing.mdx
```diff
@@ -151,7 +151,7 @@ The purpose of the operation is to ensure this checkout is correct. The validati
 
 1. The required addresses are valid.
 
-2. All selected products are in stock (while making the purchase, another user could already buy the last available item).
+2. All selected products are in stock (while making the purchase, another user could already buy the last available item), except variants that have disabled stock tracking which can have negative stock.
```
patrys, line 154: Is the negative part important here? I think it would be easier to say, "except for variants that have stock tracking disabled in which case the stock level does not matter."

timuric: I agree, but my intention was to add "negative stock" as a keyword to be discoverable via search as this is a common question

patrys: Can we then say "zero or negative?"

patrys (approved):

PR #860, opened 2023-08-05
replace curly braces in unicode with glyph char

### LINKED ISSUE #824: Fix metadata description
Use curly braces instead of #x007B;
image

qunm00:
Fix #824

### DIFF
--- a/docs/api-reference/miscellaneous/scalars/metadata.mdx
+++ b/docs/api-reference/miscellaneous/scalars/metadata.mdx
@@ -76,10 +76,10 @@ Metadata is a map of key-value pairs, both keys and values are `String`.
 
 Example:
 
-```
+```json
 {
-    "key1": "value1",
-    "key2": "value2"
+  "key1": "value1",
+  "key2": "value2"
 }
 ```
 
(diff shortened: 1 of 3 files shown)


docs/api-reference/miscellaneous/scalars/metadata.mdx
```diff
@@ -77,10 +77,10 @@ Metadata is a map of key-value pairs, both keys and values are `String`.
 Example:
 
 ```
```
mmiszy, line 79: Wouldn't it have worked if we did this instead?
```suggestion
```json
```
```

nmiquan: It doesn't work. And I think it's benenficial to replace unicode with curly braces for readability in any case.

In addition, specifying the code block as json would provide syntax highlighting, as below
image

Without specifying json, the code block would look like below
image

qunm00: I looked through the project and found that json code blocks have syntax highlighting. So I added json to the metadata code blocks.

patrys (approved):

2can (approved):

PR #861, opened 2023-08-06
move filtering from guides to api usage

### LINKED ISSUE #820: Move filtering to API usage
Move https://docs.saleor.io/docs/3.x/developer/filtering to API usage

qunm00:
Fix #820


patrys (approved):

patrys:
This will also need a redirect entry in the Docusaurus config file.

qunm00:
> This will also need a redirect entry in the Docusaurus config file.

Thanks for pointing this out. I realize that redirect is needed for the case when the page is reference elsewhere in the docs, the reference won't redirect to a dead link.

I tested the redirect in the Preview. It works fine.

PR #864, opened 2023-08-08
Add docs for bulk attributes

SzymJ:
This adds `bulk-attributes` and `error-policy` sections.

### DIFF
--- a/docs/developer/bulks/bulk-attributes.mdx
+++ b/docs/developer/bulks/bulk-attributes.mdx
@@ -1,5 +1,5 @@
 ---
-title: Bulk Attributes
+title: Bulk Attribute Import
 ---
 
 :::info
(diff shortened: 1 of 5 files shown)


docs/developer/bulks/bulk-attributes.mdx
```diff
@@ -0,0 +1,351 @@
+---
+title: Bulk Attributes
```
patrys, line 2: I don't think "Bulk Attributes" is a good title as it's the operation that happens in bulk, not the attributes. "Bulk Attribute Updates," or "Bulk Attribute Import" would be better.

The same goes for the existing "Bulk Orders" document which to me sounds more like "placing large orders" than "creating many orders at once."

maarcingebala (approved):

patrys (approved):

PR #883, opened 2023-08-18
Describe & improve api reference update process

2can:
This PR improves the way the update of API reference is done. 

After the change both `api-reference` and `storefront-reference` is updated with one command. 

The update process removes the outdated files which previously were left intact. 

The separate steps have been put into different scripts which allows it to easily replicate/run it's behavior.


README.md
```diff
@@ -207,15 +207,25 @@ For more information about the navigation bar, click [here](https://docusaurus.i
...
 To update the API reference:
 
-1. Start Saleor API at `http://localhost:8000`
+1. Put `schema.graphql` locally into the root of the saleor-docs repository
 1. Run `npm run update-api-reference`
 
+The script behind the scenes does several steps:
+
+1. The api-reference is generated in `.tmp` folder.
+1. The api-storefront is generated in `.tmp` folder. This step also includes the generation of storefront schema. Storefront schema is a subset of the Saleor schema with filtered-out elements that require permissions.
```
korycins, line 224: > with filtered-out elements that require permissions.

I know this is not the scope of this PR, but IMO we need to change this approach, and not include mutations/queries that require OWNER or AUTHENTICATED_USER. Currently, any mutation like set-default-shipping-address for customer, are placed in api-reference, instead of a storefront-api

korycins (approved):
Looks really great!

zaiste (approved):

PR #900, opened 2023-08-24
Add Payload CMS docs to cms app

lkostrowski:


docs/developer/app-store/apps/cms.mdx
```diff
@@ -160,6 +161,32 @@ Each of the CMS platforms has slight differences in configuration and behavior d
 - Fields are not autocompleted. You must ensure mapping is correct.
 - Builder.io doesn't support unique constraints on fields - hence the app will not guarantee that duplicates will not be created.
 
+### [Payload](https://payloadcms.com/)
+
+Payload CMS requires some extra steps to be performed before the app can be used.
+
+The app uses REST API, by default enabled in Payload. Typically, it's available under `/api` path.
+You need to put this URL in the configuration form.
```
peelar, line 169: **issue**:
Those two sentences aren't super clear to me.

> The app uses REST API

To me, that suggests that the app has its own REST API. I think you meant that the app communicates with Payload's REST API.

> Typically, it's available under `/api` path.

We could use some more context here. Maybe a link to the docs?

> You need to put this URL in the configuration form.

Maybe we can name a specific field in which we need to put the URL.

docs/developer/app-store/apps/cms.mdx
```diff
@@ -160,6 +161,32 @@ Each of the CMS platforms has slight differences in configuration and behavior d
...
+The user collection used for authentication has a slug (set from the coded configuration). You need this slug to be set in the "Authenticated user slug" field in the config form.
+
+Lastly, you need to create a collection for product variants. The app will create items in this collection. Please create all corresponding fields and ensure the one for _variant id_ is configured as unique.
+
+The product variant collection slug must be set in the form as well - as a collection slug field.
+
+If you want to play with the app without authorization, you need to set all access rights (read, write, create, delete) to "true" (public access) in Payload.
+Please leave the "Authenticated user slug" and "User API key" fields empty.
+
+If auth is set in the app, the app will attach the API key to the requests, and it's on you to configure Payload to accept it. Usually its enough to check if `request.user` exists.
```
peelar, line 184: **suggestion (non-blocking)**:
There are quite a lot of terms and field names thrown around in this fragment. If we can, we should link to Payload docs wherever possible.

peelar (approved):

docs/developer/app-store/apps/cms.mdx
lkostrowski: @peelar I have rewritten it all a little

peelar (approved):

2can (approved):

PR #912, opened 2023-08-28
Add migration guide for NOTIFY_USER

maarcingebala:
Docs for https://github.com/saleor/saleor/issues/13803

I'll see if I can add links to API reference for subscription types in this PR or later, I think some of the subscription types are not grouped correctly.

### DIFF
--- a/docs/upgrade-guides/notify-user-deprecation.mdx
+++ b/docs/upgrade-guides/notify-user-deprecation.mdx
@@ -1,32 +1,44 @@
 ---
-title: Migrating from USER_NOTIFY to dedicated events
+title: Migrating from NOTIFY_USER to dedicated events
 ---
 
-Starting from Saleor 3.16, the `NOTIFY_USER` webhook event will be deprecated. Apps that use this webhook should migrate to dedicated webhook events. Below is a list of webhook events that replace particular notifications. All webhook events that replace `NOTIFY_USER` support [subscription queries](developer/extending/webhooks/subscription-webhook-payloads.mdx#custom-payloads), allowing you to define the payload you want to receive for your notifications.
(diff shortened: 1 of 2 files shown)


docs/upgrade-guides/notify-user-deprecation.mdx
```diff
@@ -0,0 +1,32 @@
...
+
+Starting from Saleor 3.16, the `NOTIFY_USER` webhook event will be deprecated. Apps that use this webhook should migrate to dedicated webhook events. Below is a list of webhook events that replace particular notifications. All webhook events that replace `NOTIFY_USER` support [subscription queries](developer/extending/webhooks/subscription-webhook-payloads.mdx#custom-payloads), allowing you to define the payload you want to receive for your notifications.
+
+The table below shows sub-types of `USER_NOTIFY` events (represented as the `notify_event` field in the legacy payloads), their GraphQL subscription replacements, and the Saleor version in which they were or will be introduced.
+
+To migrate to new events, you need to create new webhooks in Saleor and provide subscription queries for the events you want to receive.
+
+| USER_NOTIFY event                | Subscription                                     | Saleor version |
+| -------------------------------- | ------------------------------------------------ | -------------- |
+| `ACCOUNT_CHANGE_EMAIL_CONFIRM`   | `AccountEmailChanged`                            | 3.15           |
```
korycins, line 13: Wouldn't it be worth including the api-reference to each subscription type?
Now you need to copy the type and search it manually, we could simplify this process.

maarcingebala: Yes, as I wrote in the description I think some subscriptions types are now in wrong groups, since they're URLs are static I would have to fix the groups first, or later fix the paths once the grouping is fixed.

Air-t (approved):

docs/upgrade-guides/notify-user-deprecation.mdx
2can: @maarcingebala please create 2 issues in docs, so we don't miss that afer merging this PR
1. for a review of  grouping of subscription types
2. for linking the `NOTIFY_USER migration guide` with api-reference

maarcingebala: Added links to API reference.

patrys (approved):

PR #926, opened 2023-09-04
Describe the search in README

2can:


README.md
```diff
@@ -255,6 +255,57 @@ The script behind the scenes does several steps:
...
+
+For even finer control over search result rankings, you can manually influence the ranking of specified pages by adding a custom meta attribute - `rank` - to the page. The `rank` meta is configured to have the highest priority in Algolia.
+
+To assign a custom rank to a particular page, use the following code snippet:
+
+```
+<head>
+  <meta name="rank" content="50" />
+</head>
+```
```
patrys, line 307: Why `<head>` instead of MDX frontmatter?

2can: How would you do that with frontmatter so it's present to the Algolia Crawler?

2can: I think there is no way to add custom meta tags through the Docusaurus frontmatter https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-content-docs#markdown-front-matter

zaiste (approved):

PR #1017, opened 2023-11-08
Improve shipping app documentation

krzysztofwolski:
- add the required permission section to the shipping app section
- add warning about circ. import in subscription query


peelar (approved):

2can:
@krzysztofwolski please also update `canary`, without that this change will be overwritten with the next Saleor release

PR #1035, opened 2023-11-22
Update 3.x docs to the latest Saleor release

2can:


zaiste (approved):

mmiszy (approved):

versioned_docs/version-3.x/upgrade-guides/3-17-to-3-18.mdx
```diff
@@ -0,0 +1,9 @@
+---
+title: Upgrading from 3.17 to 3.18
```
kadewu, line 2: Is there a reason to create this file? It's probably generic and purpose I believe would be to make sure that users follow proper zero-downtime strategy.
But then this is misleading, cause we are constantly releasing new patches, and the version should be always the last patch.

I believe it would be better to just link to: https://saleor-docs-d9j8166xb-saleorcommerce.vercel.app/docs/3.x/setup/upgrading#upgrading-with-zero-downtime if we intend with this file.

kadewu (commented):
Don't we miss sidebar update?

PR #1077, opened 2024-01-29
Always apply single promotion rule

korycins:


docs/developer/discounts/promotions.mdx
```diff
@@ -19,9 +19,16 @@ The promotion discount will be also visible on the checkout line prices after ad
 product to the card. The checkout subtotal is calculated as a sum of discounted prices of variants
 added to the cart.
 
+###### Saleor 3.17 (Deprecated)
+
 When multiple promotions apply to one product, the promotion with the greatest savings is used.
 The discounts from rules within the single promotion are summed up.
```
IKarbowiak, line 25: Should we leave in docs what was the previous behaviour? Maybe we should mention this change only in the upgrade guide, and here in the promotion page stay with the current behavior?

2can (approved):

IKarbowiak:
I think that we should also mention this change in the upgrade guide for 3.19

docs/developer/discounts/promotions.mdx
korycins: I did this in the same way as we have for transactionAPI: https://docs.saleor.io/docs/3.x/developer/payments#saleor-34-deprecated

korycins:
@IKarbowiak Done! 👮‍♀️

IKarbowiak (approved):

PR #1115, opened 2024-03-13
Restricting Saleor Cloud API access article

peelar:
**Context**:
We often get questions about how our Cloud users can restrict access to Saleor API. This article goes through the possible options and describes an example use case.

**Preview:** https://saleor-docs-git-restrict-api-access-saleorcommerce.vercel.app/docs/3.x/restricting-api-access

### DIFF
--- a/docs/restricting-api-access.mdx
+++ b/docs/restricting-api-access.mdx
@@ -9,31 +9,25 @@ Below, we explore two ways Saleor Cloud allows you to restrict access to the API
 
 ## Allowed API origins
 
-Most times, the list of the origins you expect to call your API is known in advance. By specifying the allowed origins, you can prevent unauthorized access to your API.
+If the list of the origins you expect to call your API is known in advance, you can automatically reject the unknown ones.
 
-In the project view of your Saleor Cloud dashboard, you can customize the allowed origins. The default option is to "Allow all", but you can also specify a list of origins or set it to "Dashboard only".
+In the project view of your Saleor Cloud dashboard, you can customize the allowed origins. The default option is to "Allow all" but you can also specify a list of origins or set it to "Dashboard only".
 
 :::info
 
-Changing the settings to only allow the selected origins can affect the installed apps. If you want your app to call the Saleor API from the client, you need to add the app's origin to the list of allowed origins. You can read more about app's access scopes in the [App permissions](developer/extending/apps/architecture/app-permissions.mdx#access-scopes) document.
+Changing the settings to only allow the selected origins can affect the installed apps. If you want your app to call the Saleor API from the client, you need to add the app's origin to the list of allowed origins. You can read more about the app's access scopes in the [App permissions](developer/extending/apps/architecture/app-permissions.mdx#access-scopes) document.
 
 :::
-
-Here is how an example list of allowed origins may look:
-
-- `https://your-storefront.com` - your storefront
-- `https://your-app-1.com` - your first app (if it is calling the Saleor API from the client)
-- `https://your-app-2.com` - your second app (if it is calling the Saleor API from the client)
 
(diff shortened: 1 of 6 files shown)


docs/restricting-api-access.mdx
```diff
@@ -0,0 +1,169 @@
...
+
+:::
+
+Combined with a server-side proxy, this can be a powerful tool to obscure the Saleor API from unauthorized access. We will explore this in more detail in the example below.
+
+## Example: Password-protected API proxy
+
+### Overview
+
+Let's imagine you have a storefront that communicates with Saleor API. Your storefront, naturally, implements some sort of a checkout flow. In Saleor, most checkout actions don't require the user to be authenticated. It is done so by design - you want to allow users to manage their cart without being logged in.
```
krzysztofwolski, line 46: Would rather choose to describe in detail the usecase which is implemented in the example - getting the list of products. The webhooks rant is adding unnecessary context here.

It is a valid usecase - by default no authentication is required to get the list of available products. If shop owner wants to restrict that data only for the logged in users, Password API Protection for the API is the solution.

krzysztofwolski (changes requested):

docs/restricting-api-access.mdx
```diff
@@ -0,0 +1,169 @@
...
+
+In the project view of your Saleor Cloud dashboard, you can customize the allowed origins. The default option is to "Allow all", but you can also specify a list of origins or set it to "Dashboard only".
+
+:::info
+
+Changing the settings to only allow the selected origins can affect the installed apps. If you want your app to call the Saleor API from the client, you need to add the app's origin to the list of allowed origins. You can read more about app's access scopes in the [App permissions](developer/extending/apps/architecture/app-permissions.mdx#access-scopes) document.
+
+:::
+
+Here is how an example list of allowed origins may look:
```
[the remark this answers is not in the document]
peelar: > Are wildcards (eg https://*.example.com) supported? If so, it would be good idea to mention it here.

I don't think so, the "*" character isn't even allowed by the regex.

docs/restricting-api-access.mdx
peelar: addressed

docs/restricting-api-access.mdx
```diff
@@ -0,0 +1,157 @@
...
+
+Set up "API Password Protection" in the Saleor Cloud dashboard as [described above](#api-password-protection).
+
+#### Step 2: Visit your current implementation
+
+In the current implementation, your Next.js application calls the Saleor API directly:
+
+:::note
+
+In a production-grade Next.js application, you most likely would use some GraphQL client (like [Apollo Client](https://www.apollographql.com/docs/react/)) to fetch the data from the Saleor API. However, for simplicity, we will use the native `fetch` API in this example.
```
timuric, line 58: One of our core value propositions is being "technology agnostic", providing Next.js as an example feels quite opinionated, I would recommend providing a more generic example

peelar: I see your point. Because I already rewrote the example once (based on Krzyh's feedback), I only added a little Next.js info box that explains the choice. If that creates confusion in the future, I will rewrite it.

krzysztofwolski (changes requested):

docs/restricting-api-access.mdx
krzysztofwolski: On the other hand, NextJS might be the most common tool for creating storefronts with additional BBF functionalities. This example and docs will for sure find a broad audience and should be easy to transfer to other stacks.
I think it's good as is and we can come back to it in the future

peelar:
@krzysztofwolski please check again

krzysztofwolski (approved):

2can (approved):

PR #1148, opened 2024-04-26
Remove api-storefront

2can:
Currently, the Storefront API is quite confusing. Browsing API is tricky as you need to know exactly which one you use. The search doesn't work well. At this point, removing the Storefront API is better because it's quite confusing to the API consumers.


aniav (approved):
Please remember to set the redirects from storefront URL 🙌

aniav:
We have "Admin API" in the generated Preview in multiple places. We should call it "API Reference" in the header and in other places:

Screenshot 2024-04-26 at 19 04 09

krzysztofzuraw (approved):

andrzejewsky (approved):

PR #1153, opened 2024-05-13
Flow on how to go back from C&C to standard delivery method

tomaszszymanski129:
Related PR: https://github.com/saleor/saleor/pull/15975
Ticket: https://linear.app/saleor/issue/SHOPX-568/switching-from-clickandcollect-to-shipping-automatically-pre-fills-the

### DIFF
--- a/docs/developer/checkout/address.mdx
+++ b/docs/developer/checkout/address.mdx
@@ -229,6 +229,14 @@ As a result, we get an updated checkout object with a delivery method set:
 }
 ```
 
+### Switching from C&C to standard delivery method
+
+After choosing the click & collect option on a checkout, going back to a standard delivery method will require the shipping address to be set, this can't be done directly by checkoutDeliveryMethodUpdate, but can be achieved using the flow below.
+
+1. First use `checkoutDeliveryMethodUpdate` (just with `id` or `token` provided). This will remove the collection point from checkout and also remove the shipping address.
+2. Now `checkoutAddressUpdate` can be used to set the new shipping address for checkout.
+3. Finally, use `checkoutDeliveryMethodUpdate` with `deliveryMethodId` to set the standard delivery method.
+
 ## Address Validation
 
 The checkout's mutations that accept an address as an input have a field that can turn off the address validation. It allows assigning a partial or not fully valid address to the checkout. Providing country code is mandatory for all addresses regardless of the rules provided in this input.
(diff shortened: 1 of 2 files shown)


aniav (approved):

docs/developer/checkout/address.mdx
```diff
@@ -229,6 +229,14 @@ As a result, we get an updated checkout object with a delivery method set:
 }
 ```
 
+### Switching from C&C to standard delivery method
+
+After choosing the click & collect option on a checkout, going back to a standard delivery method will require the shipping address to be set, this can't be done directly by checkoutDeliveryMethodUpdate, but can be achieved using the flow below.
```
IKarbowiak, line 234: This sounds for me a little bit like we have a bug in our code. IMO we should describe it as an expected flow, and maybe add a information why the shipping address cannot be just changed. 
- `going back to a standard delivery method will require the shipping address to be set` - the checkout already have the shipping address set but it points to CC, so it need to be changed not set
- `going back to a standard delivery method will require the shipping address to be set, this can't be done directly by checkoutDeliveryMethodUpdate` - but the address is not updated by `checkoutDeliveryMethodUpdate` mutation, so this sentence is misleading

docs/developer/checkout/address.mdx
```diff
@@ -229,6 +229,14 @@ As a result, we get an updated checkout object with a delivery method set:
...
+
+1. First use `checkoutDeliveryMethodUpdate` (just with `id` or `token` provided). This will remove the collection point from checkout and also remove the shipping address.
```
IKarbowiak, line 236: IMO:
- we should add link to the API docs for this mutation: https://docs.saleor.io/docs/3.x/api-reference/checkout/mutations/checkout-delivery-method-update
- we should add graphl example for that like:
```
mutation {
  checkoutDeliveryMethodUpdate(id: "your_checkout_id") {
    checkout {
      id
    }
  }
}
```

docs/developer/checkout/address.mdx
```diff
@@ -229,6 +229,14 @@ As a result, we get an updated checkout object with a delivery method set:
...
+2. Now `checkoutAddressUpdate` can be used to set the new shipping address for checkout.
```
IKarbowiak, line 237: It's not a migration that should be used. The valid migration is `checkoutShippingAddressUpdate`. We should also add a link to this migration

szczecha: Mutation 😄 but I agree with Iga's

IKarbowiak (approved):

Issue #1058, opened 2023-12-14
Add info about which country is used for tax calculation in case of different addresses


szczecha:
We are missing information in the docs on which country it takes to calculate taxes it can be misleading for checkout with digital products when customer pass the shipping address at checkoutCreate mutation.

The shipping address is always set when supplied in the input.

The order of resolving the checkout country code is always as follows:
Shipping address
Billing address
The channel's default country

https://www.notion.so/saleor/Tax-addresses-c831a50c95764e78a7e4a30798a2cc3d

PR #1156, opened 2024-05-14
Update the tax address hierarchy docs

szczecha:
Fix https://github.com/saleor/saleor-docs/issues/1058

### DIFF
--- a/docs/developer/taxes.mdx
+++ b/docs/developer/taxes.mdx
@@ -402,6 +402,10 @@ The process of determining the country code follows this sequence:
 2. Billing address
 3. The channel's default country
 
+:::note
+The table provides examples to help understand how tax calculation is determined based on the address.
+:::
+
 |             | Store Address | Shipping Address | Billing Address | Warehouse Address |
 | ----------- | ------------- | ---------------- | --------------- | ----------------- |
 | Country     | US            | US               | US              | US                |
@@ -409,13 +413,13 @@ The process of determining the country code follows this sequence:
 
 ### Example for shipable products
 
-- Create a checkout with channel and lines. Since the shipping address is empty, taxes are calculated based on the store address (Country US, state LA)
+- Create a checkout with channel and lines. Since the shipping and billing addresses are empty, taxes are calculated based on the store address (Country US, state LA)
 - Update the billing address. The shipping address is still empty, but since the billing address is now available, taxes are calculated based on it (Country US, state TX)
 - Update the shipping address. Taxes are now calculated based on the shipping address (Country US, state VA).
 
 ### Example for digital products
 
-- Create a checkout with channel and non-shippable lines. Since the shipping address is empty, taxes are calculated based on the store address (Country US, state LA)
+- Create a checkout with channel and non-shippable lines. Since the shipping and billing addresses are empty, taxes are calculated based on the store address (Country US, state LA)
 - Update the billing address. Taxes are calculated based on it (Country US, state TX)
 
 :::warning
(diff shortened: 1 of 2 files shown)


docs/developer/taxes.mdx
```diff
@@ -394,3 +393,38 @@ mutation {
...
+
+## Address for tax calculation
+
+The process of determining the country code follows this sequence:
+
+1. Shipping address
+2. Billing address
+3. The channel's default country
+
+|             | Store Address | Shipping Address | Billing Address | Warehouse Address |
```
korycins, line 405: I understand the sequence above, but I don't know what the table represents.  Does it represent that each address can have different country? Maybe worth adding some description or something?

docs/developer/taxes.mdx
```diff
@@ -394,3 +393,38 @@ mutation {
...
+- Update the billing address. The shipping address is still empty, but since the billing address is now available, taxes are calculated based on it (Country US, state TX)
+- Update the shipping address. Taxes are now calculated based on the shipping address (Country US, state VA).
+
+### Example for digital products
+
+- Create a checkout with channel and non-shippable lines. Since the shipping address is empty, taxes are calculated based on the store address (Country US, state LA)
+- Update the billing address. Taxes are calculated based on it (Country US, state TX)
+
+:::warning
+It's possible to set shipping address in the checkoutCreate mutation. In this case, the shipping address will be used for tax calculation since we always prioritize shipping address over billing address.
```
korycins, line 422: Isn't it a bug that we don't clean up the shipping address when we don't have anything to send?
I would say it is a bug, as some mutations can raise `SHIPPING_NOT_REQUIRED` error.

docs/developer/taxes.mdx
korycins: Ohh, it is clear after checking the examples. Maybe worth adding a note to the table, that this is a helper table for examples?

docs/developer/taxes.mdx
szczecha: @korycins We change it in Saleor 3.15 https://github.com/saleor/saleor/blob/719d6c0da615161db8669a4626a90d6ebc45ccd3/CHANGELOG.md?plain=1#L350 


Until now, checkout mutations were ignoring provided shipping address when shipping was not required. After this change, the shipping address is always set when supplied in the input. It might be breaking, as the shipping address affects the country code used for tax calculation.
The order of resolving the checkout country code is always as follows:
Shipping address
Billing address
Channel's default country

docs/developer/taxes.mdx
szczecha: note added 625f505696e2c6ae47d38f3d0b496a32ad451d2f

IKarbowiak (approved):

docs/developer/taxes.mdx
korycins: @szczecha I mean: You're writing, that in some cases (like checkoutCreate) we can set the shippingAddress, but later you will not be able to call `CheckoutShippingAddressUpdate`, as this mutation will raise an exception `SHIPPING_NOT_REQUIRED`. So it would be confusing for me, that in some cases, you can set shippingAddress, but with the same data, you will not be able to call another mutation, that will do the part of checkoutCreate work.

korycins (approved):

docs/developer/taxes.mdx
szczecha: I created an issue https://linear.app/saleor/issue/SHOPX-730/allow-to-set-shipping-address-for-digital-checkouts-in, when someone fixes it will also update the documentation

lkostrowski (approved):

PR #1166, opened 2024-05-28
Update PostgreSQL version in Windows installation instructions

NyanKiyoshi:
This changes the PostgreSQL version from v12 to v15 in the Windows manual (non-containerized) installation instructions as we no longer actively maintain v12, only v15.

Acts as a follow-up to:
- https://github.com/saleor/saleor-platform/pull/309
- https://github.com/saleor/saleor/pull/16055

### DIFF
--- a/docs/setup/windows.mdx
+++ b/docs/setup/windows.mdx
@@ -24,7 +24,7 @@ Saleor requires **Node.js 10.0** or later. Go to the [Node.js downloads page](ht
 
 ### PostgreSQL
 
-You will need **PostgreSQL** version **12.x**. Get the Windows installer from the [project’s download page](https://www.postgresql.org/download/windows/).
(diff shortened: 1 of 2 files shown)


patrys (approved):

aniav:
Will this change be ported to older versions of `core`? I'm asking to make sure we shouldn't also add it in `versioned_docs`.

patrys:
It's safe to use Postgres 15 with all supported versions of Saleor. However, I do not plan to port the related dev container change as it would force everyone to deal with the manual DB upgrade process.

aniav:
I am asking only about this PR to the docs. If it can be used in e.g. 3.19 we have to add the same changes in `versioned_docs` directory.

patrys:
It can be used with 3.14+. The previous version told people to use Postgres 12 on Windows, but I'm not even sure that will work, as we've all been using Postgres 13 for ages.

aniav (approved):

aniav:
Will be merged when the other PRs are merged.

PR #1173, opened 2024-06-07
Add channels

timuric:
Intentionally omitted copying to versioned docs to make review process more simple

PR #1180, opened 2024-06-12
Multichannel split

timuric:


aniav:
#1173 is merged so you should be unblocked on this one 🙌

aniav (approved):
I added one missing redirect in vercel.json. Aside from that I think it's good to be merged.

There are places where the content should be improved to make it consistent with/without infinitives, but we will do it in another round of changes to not block this.

PR #1187, opened 2024-06-18
Structure overhaul

patrys:
- Split documentation into several pages with separate sidebars
- Unified the structure somewhat
- Removed what looks like accidental commits to the API reference
- Got rid of some complex custom sidebar formatting
- Dropped giscus and "was this page helpful?"


src/css/components/menu.css
```diff
@@ -15,28 +15,10 @@
 
 .menu__link {
   font-size: var(--font-size-sm);
-  color: var(--gray10);
   transition: all 0.3s ease-in-out;
```
peelar, line 18: **suggestion**:
(I hope this is the right place)

IMO, we should decrease the duration of the transition. Moving cursor from top of the sidebar item list to the bottom leaves a background color "smudge" for too long (to my taste).


https://github.com/saleor/saleor-docs/assets/44495184/6cf4319b-57a1-4b97-9a46-6c6b745785ba

peelar (commented):
- "Core Concepts" is missing an overview page
image

- "Core Concepts" has its item in both the sidebar navigation and top navigation
Zrzut ekranu 2024-06-19 o 12 42 50

src/css/components/menu.css
patrys: This is completely unrelated to this PR, though. I only reverted even uglier sidebar styling to its original state.

timuric: Current docs don't have group pages, so we would need to solve it somehow

patrys:
@peelar 

>"Core Concepts" is missing an overview page

You are welcome to write something there. I wanted to have a predictable link target but I don't have the time to create new content.

>"Core Concepts" has its item in both the sidebar navigation and top navigation

That's intentional. The same is true for all the other links from the top nav as they are merely shortcuts.

timuric:
I think "Extending Saleor" belongs to core concept

src/css/components/menu.css
peelar: I disagree it's unrelated. You changed the way the sidebar looks and I was just sharing my opinion that this particular bit, to me, now looks off.

timuric:
About Saleor can not be collapsed as Why saleor
image

src/css/components/menu.css
patrys: I removed all easing from the sidebar, so the animation is gone entirely.

patrys:
@timuric

>I think "Extending Saleor" belongs to core concept

I wanted "core concepts" to be about using Saleor, not building stuff around it. You are welcome to propose a better name for the section. Medusa calls a similar thing "Commerce Modules."

maarcingebala (approved):

aniav (approved):

timuric:
@aniav we can merge now?

aniav:
@timuric had to do path tweaks, once vercel builds I will merge to not block anything any more 🙌

PR #1204, opened 2024-07-03
Redesign

timuric:


peelar:
image

**suggestion**: We have "Apps" at the top level of "Customizing", and then inside "Extending" we have "Apps" again. Given that the first one is the link to the App Store section of the docs, I think it should be called "App Store".

Also, once we rename it, I think we could get rid of the "Extending" catalog and unpack its content in the "Customizing" section.

peelar:
image

TBH, I am not sure how I feel about removing the top level nav items. I understand the items are now nested in the left sidebar, but for me the experience is a bit inconsistent. There is nothing about the "Core Concepts" or "Apps" sidebar item that suggests there is a lot of content hiding under that label. I think I would leave "Core Concepts", "API Reference" and "App Store" in the top level nav and just remove it from the sidebar.

timuric:
@peelar the idea is that there is no longer going to be such nested structure, so this section will be split differently

sidebars.js
```diff
@@ -1,15 +1,29 @@
...
+    {
+      type: "doc",
+      id: "index",
+      label: "Overview",
+    },
     {
       type: "category",
-      label: "About Saleor",
-      collapsible: false,
+      label: "Quick Start",
```
peelar, line 21: **style**: I would go with "Quickstart"

components/Steps/index.jsx
peelar, line 1: Awesome component 😍

timuric:
@peelar sorry I forgot to mark this PR as draft as this is WIP

peelar:
> @peelar sorry I forgot to mark this PR as draft as this is WIP

Ah, got it!

docs/upgrade-guides/GUIDELINES.mdx
peelar, line 1: **suggestion (non-blocking)**: This is not supposed to be visible in the sidebar (as far as I can see, it is not). I think we can remove this file altogether, this is for internal use only, and we maintain this type of content outside our docs.

peelar (changes requested):
image

1. **style (non-blocking)**: "Extending" -> "Webhooks"

image

2. **style (non-blocking)**: IMO, the "quickstart" in "Cloud quickstart" is redundant here.

3. **suggestion (blocking)**: I think App Store is missing from the sidebar.

4. **thought**: Maybe we could add a new article about App Store to "Why Saleor"? 🤔 Not in this PR, I am just thinking out loud.

krzysztofwolski (commented):
I'm findind myself confused which section I'm in ATM.

image

From the first look its hard to know thats 'building apps' section. Could be mitigated with displaying name of the section in the sidebar or breadcrumbs. 

Besides that im enjoying a new look :+1:

timuric:
@peelar on one hand Extending section is mostly about webhooks but there is also an overview that generally explains extensibility, imo we should evolve that section beyond webhooks, for example describing middleware pattern, automation etc

timuric:
@krzysztofwolski I am planning to add titles with icons to each sidebar section, but this PR is becoming too broad, so I will address it in a smaller follow up refinement

docs/upgrade-guides/GUIDELINES.mdx
timuric: done

timuric:
@peelar agree about removing kickstart but unfortunately there are barely any docs in Cloud section, so I used it as a link there, which would be confusing without a full label, this is a temporary measure until we write more docs for cloud 
image

peelar (approved):

PR #1213, opened 2024-07-05
Add sidebar titles

timuric:
Added titles to chapters
image
image


peelar:
IMO, this title looks too much like a button. I want to click it, it even has a cursor: pointer. Can we remove the border/shadow and probably some padding? Adding the icon is enough for me.

timuric:
It doesn't work that well without special treatment, will have to think about  
image
image

timuric:
@peelar change the style, no longer looks like a button 
image

image

peelar (approved):

PR #1214, opened 2024-07-08
Extend pricing page with the amount precision

IKarbowiak:
Make it clear what is the precision of returned prices.

### DIFF
--- a/docs/api-usage/prices.mdx
+++ b/docs/api-usage/prices.mdx
@@ -3,7 +3,7 @@ title: Prices
 ---
 
 Saleor API has four different types that represent prices. The most basic type consists of a currency code and a numeric amount. The most complex one can define a range of prices, including net, gross, and tax amounts.
-The numeric amount precision is determined by the precision of the price currency. For example, USD has a precision value of 2, such as $1.00
+The numeric amount precision is determined by the precision of the price currency. For example, USD has a precision value of 2, such as $1.00.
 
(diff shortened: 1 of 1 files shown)


lkostrowski (approved):

docs/api-usage/prices.mdx
```diff
@@ -3,6 +3,7 @@ title: Prices
 ---
 
 Saleor API has four different types that represent prices. The most basic type consists of a currency code and a numeric amount. The most complex one can define a range of prices, including net, gross, and tax amounts.
+The numeric amount precision is determined by the precision of the price currency.
```
timuric, line 6: I would suggest adding an example:

 The numeric amount precision is determined by the precision of the price currency. For example, USD has a precision value of 2, such as $1.00

docs/api-usage/prices.mdx
```diff
@@ -3,6 +3,7 @@ title: Prices
 ---
 
 Saleor API has four different types that represent prices. The most basic type consists of a currency code and a numeric amount. The most complex one can define a range of prices, including net, gross, and tax amounts.
+The numeric amount precision is determined by the precision of the price currency. For example, USD has a precision value of 2, such as $1.00
```
peelar, line 6:
```suggestion
The numeric amount precision is determined by the precision of the price currency. For example, USD has a precision value of 2, such as $1.00.
```

peelar (changes requested):

timuric (approved):

peelar (approved):

IKarbowiak:
Thank you guys for taking care of that! ❤️  🚀

PR #1216, opened 2024-07-09
FIX typo

teacherSsamko:
port 9000 is for dashboard. you need to use 8000 port for graphql


timuric (approved):

timuric:
@teacherSsamko thank you for correcting it!

PR #1227, opened 2024-07-16
Schedule API reference updates

timuric:
Run every tuesday


2can:
Closing, the workflow uses versioning which has been removed. The workflow logic needs to be refactored.

PR #1229, opened 2024-07-16
Update workflow

timuric:


2can:
Closing, the workflow uses versioning which has been removed. The workflow logic needs to be refactored.

PR #1240, opened 2024-07-17
Set docs as routeBasePath

2can:
This PR sets docs-only mode which removes `docs/` from the pathname.


peelar (approved):

krzysztofwolski (approved):
Seems to be working all right:
- tested redirection from old /docs 
- tried links in docs
- edit page button works

PR #1248, opened 2024-07-23
Restructuring payment docs

timuric:
* Split into chapters for easier navigation and digestion
* New overview guide to make it easier to make a decision where to get started
* Lifecycle chapter to unify few points
* Removed feature preview from transactions API, which some of them dated as long as 3.4

The custom app angle is a bit too specific as it assumes a certain angle of payments being used via storefront, but in reality transaction apis can be used in other context, thus small change there


peelar (dismissed):

docs/developer/payments/deprecated.mdx
```diff
@@ -0,0 +1,75 @@
...
+    Dashboard -->>- Staff: Triggered
+    Saleor ->>+ App: Send webhook<br>transaction-action-request<br> with all details related<br> to action and transaction
+    App -->>- Saleor: OK
+    App ->>+ PaymentProvider: Execute refund <br>action
+    PaymentProvider -->>- App: Refunded
+    App ->>+ Saleor: Call transactionUpdate <br>to update transaction with<br> a result received after <br>requested action
+    Saleor -->>- App: Updated transaction
+```
+
+### Creating transactions in Saleor 3.4
```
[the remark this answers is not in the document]
korycins: `transactionCreate` / `transactionUpdate` are mutations that you will use, to cover more advanced cases, where you can't use transactionInitialize. It is also used on dashboard to create "manual transaction"

docs/developer/payments/deprecated.mdx
```diff
@@ -0,0 +1,75 @@
+---
+title: Deprecated (Legacy)
+sidebar_label: Deprecated
+---
+
+## Payment Plugin
+
+Using PaymentPlugin is explained in [Checkout](/developer/checkout/finalizing.mdx) section.
```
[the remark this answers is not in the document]
korycins: > Using PaymentPlugin is explained in

Also, I don't think that  we explain it there. Each plugin has different flow of handling the payment. This is why we have different docs for StripePlugin: https://saleor-docs-git-payments-restructure-saleorcommerce.vercel.app/developer/app-store/legacy-plugins/stripe  or https://saleor-docs-git-payments-restructure-saleorcommerce.vercel.app/developer/app-store/legacy-plugins/adyen

docs/developer/payments/deprecated.mdx
```diff
@@ -0,0 +1,75 @@
...
+
+### Actions Saleor 3.4
+
+If your app should also process payment request actions (like charge, refund, or CANCEL) triggered by staff users on the Saleor dashboard side or by the app,
```
korycins, line 12:
```suggestion
If your app should also process payment request actions (like charge, refund, or cancel) triggered by staff users on the Saleor dashboard side or by the app,
```

docs/developer/payments/lifecycle.mdx
```diff
@@ -0,0 +1,316 @@
...
+- the `order.totalGrantedRefund` value
+- the `order.total` value
+
+The difference between `total` and `totalGrantedRefund` is compared with **charged** transactions, and the result is one of the statuses: `PARTIAL`, `FULL`, `OVERCHARGED`. If no transaction has been charged, the status is `NONE`.
+
+#### Available transitions
+
+- Status is updated on operations that change the order's `total`, `totalGrantedRefund`, or transactions.
+
+## Recalculations of amounts
```
korycins, line 117: Maybe worth to highlight, that this part is only related to transactionAPI?

docs/developer/payments/lifecycle.mdx
```diff
@@ -0,0 +1,316 @@
...
+| :---: | :-------------------: | :----------: | :-----------------------: | :----: | :-------------------------: | :-------------------------------: | :----------------------------: |
+|   1   | AUTHORIZATION_SUCCESS |     AB12     | 2022-03-28T12:50:33+00:00 |   10   |              0              |                 0                 |               10               |
+|   2   |    CHARGE_SUCCESS     |     YZ13     | 2022-03-28T12:51:33+00:00 |   3    |              3              |                 0                 |               7                |
+
+The `AUTHORIZATION_SUCCESS` event is attached to the transaction, and its amount is added to `transaction.authorizedAmount`. Similarly, the `CHARGE_SUCCESS` event is attached to the transaction,
+and its amount is added to `transaction.chargedAmount`. Finally, the amount of the `AUTHORIZATION_SUCCESS` event is subtracted from the `transaction.authorizedAmount`.
+
+---
+
+## Event types
```
korycins, line 227: Same comment as above. It is only related to TransactionAPI

docs/developer/payments/overview.mdx
```diff
@@ -0,0 +1,43 @@
...
+
+There are few ways to work with payments in Saleor:
+
+### Payment apps
+
+Payment apps allow to process payments with external payment providers, without communicating with them directly, which is specially convenient for front-end development.
+
+- Using [payment app](/developer/payments/payment-apps.mdx).
+
+- Guide to [building payment apps](/developer/extending/apps/building-payment-app.mdx).
```
korycins, line 14: IMHO, the second link is a "part" of `Using [payment app]`. As a result of the guide you will get the payment app which uses the API explained in `(/developer/payments/payment-apps.mdx).`
Second option of building an app is making it fully custom. By saying fully custom I mean a case (only as a example), where you host own checkout flow, and provide Saleor only a results. In that case you will use mutations like: `transactionCreate`, `transactionUpdate` and `orderCreateFromCheckout`.
Ref: https://docs.saleor.io/developer/payments#what-is-a-custom-app

docs/developer/payments/overview.mdx
```diff
@@ -0,0 +1,43 @@
...
+
+- Payment plugins are included in core, so it is discouraged to modify them.
+- Building your own payment plugins is discouraged, and relies on undocumented APIs which will be deprecated.
+- Payment plugins are not as flexible as payment apps, and do not support multiple transactions on a single order.
+
+**Should you use payment plugins?**
+
+In case you need to use a payment provider that is not [available as a payment app](developer/app-store/apps/overview.mdx#built-in-apps-catalog), you can use a payment plugin. In case such provider will be available in the future swapping plugin with app would require refactoring of the checkout flow.
+
+### Payment APIs via Backend
```
korycins, line 34: Based on the context it is like "more advanced use case". Maybe worth considering a different name for this section? I totally didn't know what it was at the beginning - something like "advanced use case"?.

Also as far as I understand, from the description this covers more advanced use cases for handling a payments. I believe in that case, we should revert the `transactionCreate` from deprecated section: https://github.com/saleor/saleor-docs/pull/1248/files#r1688280173 as `transactionCreate`/`transcationUpdate` are the mutations which gives you a flexibility to cover such cases.

docs/developer/payments/deprecated.mdx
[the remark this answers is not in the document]
timuric: @korycins do we even need to document 3.4 differences? It is quite old and the full path is probably not documented anyway?

docs/developer/payments/deprecated.mdx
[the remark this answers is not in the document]
timuric: This PR is mostly about structure, but these are good points to address in the follow up PR

docs/developer/payments/lifecycle.mdx
timuric: Added `Transactions` to the title

docs/developer/payments/lifecycle.mdx
timuric: Added Transactions to the title

docs/developer/payments/deprecated.mdx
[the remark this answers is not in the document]
timuric: This is the section that I moved to a separate document, to avoid cluttering the guide with quite old examples. https://docs.saleor.io/developer/payments#saleor-34-deprecated

I suggest to remove it completely

docs/developer/payments/deprecated.mdx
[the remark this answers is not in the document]
timuric: this is how it is in old docs, and I think it is not that helpful to say it is legacy https://docs.saleor.io/developer/payments#payment-plugin-legacy

timuric:
@peelar @korycins I would like to add extra context to this PR, my intend to make only minimal adjustments to the content to  change the structure. So later when we work on content changes we can also see the delta from the original content. Otherwise if we push content and structure in the same PR it will be difficult to comprehend

docs/developer/payments/overview.mdx
timuric: I have a bit of problem with definition of Custom App vs Payment App, imo these are the same things is just one is using some webhooks and the other are not, but in reality this can be a single app that does both things. I think we need to rethink the overall grouping of terms. Probably the guide should be split between using payments vs building payments

peelar (dismissed):

timuric:
@peelar @korycins I have rewritten overview page that hopefully adds more guidance and clarity

docs/developer/payments/overview.mdx
timuric: @korycins could you please take a look at a new overview page, perhaps it is less confusing now

docs/developer/payments/deprecated.mdx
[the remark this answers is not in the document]
timuric: @korycins I have removed deprecated section

docs/developer/payments/deprecated.mdx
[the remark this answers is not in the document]
timuric: We should not call stripe as deprecated, so I will remove that in the stripe page, and this page is gone now

korycins (approved):

PR #1275, opened 2024-08-06
Update payment refunds page

IKarbowiak:
Extend the payment refunds page with `Total remaining grant calculations` section.

Related to https://github.com/saleor/saleor/pull/16486

### DIFF
--- a/docs/developer/payments/refunds.mdx
+++ b/docs/developer/payments/refunds.mdx
@@ -248,19 +248,24 @@ where
   overchargedAmount = processedAmount - order.total
   ```
 
-Below is an example order with multiple transactions and `totalRemainingGrant` on each of the stage:
+Below is an example order with multiple transactions and `totalRemainingGrant` at each stage:
 
-1. An order has a total of 100 USD and two `TransactionItem`s, one with a `chargedAmount` of 100 USD, and the second with `chargedAmount` of 60 USD.
+1. An order has a total of 100 USD and two `TransactionItem`s: one with a `chargedAmount` of 100 USD and the second with `chargedAmount` of 60 USD.
    The `totalBalance` would be 60, as `chargedAmount` - `order.total` results in 60. The `chargeStatus` is `OVERCHARGED`.
-2. Adding a granted refund with an amount of 10 USD would increase in a `totalBalance` of 10.
-   The `totalRemainingGrant` will be `10`. As `totalRefundedAmount` is 0.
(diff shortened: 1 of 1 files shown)


docs/developer/payments/refunds.mdx
```diff
@@ -204,3 +211,78 @@ mutation {
...
+  processedAmount = sum(amountCharged + amountRefunded + amountAuthorized + amountChargePending + amountRefundPending + amountAuthorizePending)
+
+  overchargedAmount = processedAmount - order.total
+  ```
+
+Below is an example order with multiple transactions and `totalRemainingGrant` at each stage:
+
+1. An order has a total of 100 USD and two `TransactionItem`s: one with a `chargedAmount` of 100 USD and the second with `chargedAmount` of 60 USD.
+   The `totalBalance` would be 60, as `chargedAmount` - `order.total` results in 60. The `chargeStatus` is `OVERCHARGED`.
+2. Adding a granted refund of 10 USD would increase in a `totalBalance` of 10.
```
korycins, line 255: You could mention here, why `totalBalance` is increased. In 1st step you say:
> The `totalBalance` would be 60, as `chargedAmount` - `order.total` results in 60

So I think worth mention that value of grantedRefund has impact on this calculation.

docs/developer/payments/refunds.mdx
```diff
@@ -204,3 +211,78 @@ mutation {
...
+| 5       | 100   | 0            | FULL            | FULL         | 90                 | 70                  | 10                        | 0                          |
+
+> **Note:**
+>
+> - `totalChargedAmount` - is the sum of charged amounts from all transactions
+> - `totalRefundedAmount` - is the sum of all refunded or pending refunds amount from all transactions
+
+:::info
+
+- Pending amounts are treated as they were processed, so are included in all processed amounts.
```
korycins, line 285: Maybe worth emphasizing, that this `info` is related to calculating the granted refund remaining amount. I mean, for example order has `totalCharged` field which doesn't include the pendingChargedAmount. Same for authorizedAmount. So maybe it would be good to explicitly write that we are talking about granted refund (yes, I am aware that we are in refund chapter, but still I was confused for a few seconds). 
It should be enough to add something like: `In case of calculating total remaining grant:`

IKarbowiak: Updated

korycins (approved):

PR #1278, opened 2024-08-07
Add linting instructions

timuric:


.husky/pre-commit
```diff
@@ -1,5 +1,4 @@
 #!/usr/bin/env sh
 
 npx pretty-quick --staged
-staged=$(git diff --name-only -- '*.mdx')
-npx eslint $staged
+npm run lint
```
krzysztofwolski, line 4: Linting should also be enabled on the CI - currently i only see "npx pretty-quick" in the buildstep.

- linting should not block build (typos should not prevent attempts in preparing preview env or deployment)
- some of the patches are submitted via GH UI. Linter output should be visible in the PR so we can fix it during the review.

CI is most likely outside of this PR scope, so this CR is optional.

krzysztofwolski (approved):
Is there

.husky/pre-commit
timuric: linter is only outputting warnings it is non blocking, it serves as a reminder after the commit, because some of the errors in front matter should be ignored

PR #1287, opened 2024-08-13
Remove example apps from docs

poulch:


timuric:
Hey @poulch please resolve the conflict so we can merge it

PR #1289, opened 2024-08-14
Checkout refactoring

timuric:


korycins (dismissed):

docs/developer/checkout/api-guide.mdx
```diff
@@ -430,3 +433,248 @@ mutation {
...
+        ...AddressFragment
+      }
+      billingAddress {
+        ...AddressFragment
+      }
+    }
+  }
+}
+
+fragment AddressFragment on Address {
```
korycins, line 467: Missing intendation:
image

docs/developer/checkout/api-guide.mdx
```diff
@@ -430,3 +433,248 @@ mutation {
...
+        "status": "UNFULFILLED"
+      },
+      "errors": []
+    }
+  }
+}
+```
+</TabItem>
+</Tabs>
+</TwoColumn>
```
korycins, line 680: Not sure if this is scope of this PR, but I think it would be worth to also explain mutations like `checkoutPromoCodeAdd`, `checkoutCustomerAttach/Detach` etc

sidebars/core-concepts.js
```diff
@@ -30,7 +30,7 @@ export const coreConcepts = [
   "developer/checkout/api-guide",
   "developer/checkout/address",
   "developer/checkout/problems",
-  "developer/checkout/finalizing",
+  "developer/checkout/cookbook",
```
korycins, line 33: Probably also not a scope of these changes, but shouldn't we have api-guide at the end of the section? 
image
We have api-guide for products and attributes at the end of the section, so I assume that the same thing should be for checkout part

docs/developer/checkout/problems.mdx
```diff
@@ -4,19 +4,6 @@ title: Problems
...
 
-:::info
-
-This feature was introduced in **Saleor 3.15**.
-
-:::
-
-:::caution
-
-This feature is in the **Feature Preview** stage, which means that it is available for experimentation and
```
korycins, line 15: Shouldn't we do this in core first? Right now we will have mismatch between docs, and `API-references` generated based on the core's schema.

In case of checkout.problems, not all potential error events are covered, this is why we keep this in preview 🤔  (at least this is my understanding).

docs/developer/checkout/problems.mdx
```diff
@@ -4,19 +4,6 @@ title: Problems
 
 ## Introduction
 
-:::info
-
-This feature was introduced in **Saleor 3.15**.
```
korycins, line 9: Is it expected that we will remove the note about the added version?

docs/developer/payments/stored-payments.mdx
```diff
@@ -11,19 +11,6 @@ Saleor uses synchronous webhooks to notify the Payment App about actions related
 
 :::info
 
-This feature was introduced in **Saleor 3.15**.
-
-:::
-
-:::caution
-
-This feature is in the **Feature Preview** stage, which means that it is available for experimentation and
```
korycins, line 20: Same concerns as for checkout.problems.

The interface was delivered, but I would get the feedback first from app developers who were using the storedPaymentMethod API, before marking this as stable.

docs/developer/checkout/problems.mdx
timuric: The issue is that only selected features have such labels; if we were to be consistent, then all of the docs would be full of such notes, which is not scalable. 

Perhaps for recent 1-2 version backs, it might be useful, but documenting 5 versions back is quite redundant, IMO. Alternatively, we can add docs release labels for anyone who wants to track back docs.

korycins (dismissed):

docs/developer/checkout/api-guide.mdx
timuric: Added extra block 
image

PR #1303, opened 2024-08-21
Remove email & messages app from docs

poulch:


lkostrowski (approved):

docs/developer/app-store/legacy-plugins/user-emails.mdx
```diff
@@ -5,7 +5,7 @@ title: User Emails
 :::warning
 This plugin is deprecated!
 
-If you plan on building a new integration with Saleor, we recommend using the [Emails and messages](../apps/emails-and-messages/overview) app instead.
+If you plan on building a new integration with Saleor, we recommend using the [SMTP](../apps/smtp/overview) app instead.
```
lkostrowski, line 8: you can also link to sendgrid app example repo

korycins:
In app's overview, there is:
```
If you plan to use SMTP server, please use the [SMTP app](https://docs.saleor.io/developer/app-store/apps/smtp/overview) 
instead. Emails and Messages app is being split into separate SMTP and Sendgrid apps.
```
But what about sendgrid? Do we have a replacement for that? I don't see any section about Sendgrid. If I am using sengrid provider, what app should I use? 

Also, wouldn't be better to leave the section for `Email and Messages` with big warning label (including side-bar) that the app is depreceated/abandoned/not available on the market anymore, and you should move to App Smtp for Smtp provider and to App X for Sendgrid provider?

krzysztofwolski (changes requested):
Maciek summed it up perfectly.
The app has users, it was mentioned in a few places so there will be people expecting to find information about it in the docs.
Removing info about it from docs adds to the confusion, not resolve it. Wait till migration is over and then remove it from docs.

PR #1319, opened 2024-09-13
Adjust information about valid order of transaction event report

IKarbowiak:
Adjust information about valid order of transaction event report.

### DIFF
--- a/docs/developer/payments/lifecycle.mdx
+++ b/docs/developer/payments/lifecycle.mdx
@@ -345,6 +345,7 @@ But to maintain accuracy and integrity, it is important that events are triggere
 The chart below illustrates the possible sequence of events that can follow one another.
 Failure to adhere to these guidelines may result in negative values and data inconsistencies.
 
+- Transaction flow strategy: `Authorization`
 ```mermaid
 stateDiagram-v2
  direction LR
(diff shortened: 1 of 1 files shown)


docs/developer/payments/lifecycle.mdx
```diff
@@ -274,6 +278,14 @@ be used to provide the history of `transactionItem` actions.
 The provided `amount` will be used to reduce `transaction.chargedAmount`
 (`transaction.chargedAmount -= amount`).
 
+### Refund events
+
+:::warning
+Refund events should only be triggered if the transaction has been previously charged.
+Triggering a refund for uncharged transactions can lead to negative `chargedAmount`.
```
lkostrowski, line 285: Is it "legal" operation? What happens if we have negative chargeAmount? Does it make sense to refund more money than was charged initially?

kadewu (dismissed):

fowczarek (approved):

docs/developer/payments/lifecycle.mdx
```diff
@@ -314,3 +334,39 @@ for `transaction.canceledAmount` will not take into account the amount from `CAN
 If an `CANCEL_FAILURE` event is related to an `CANCEL_REQUEST` event by `pspReference`, then the `transactionItem.cancelPendingAmount`
 will be reduced by the amount reported in the `CANCEL_FAILURE` event. If there is no related `CANCEL_REQUEST`, the failure event will only
 be used to provide the history of `transactionItem` actions.
+
+
+## Transaction event flow
+
+Transaction events modify the amount fields, which define the state of the transaction.
+We do not require any specific order of up comming events, the transaction amounts will be calculated
```
Magdalena-Kobylanska, line 342: "We do not require any specific order of up comming events"
should be:
"We do not require any specific order of incoming events"

Magdalena-Kobylanska (approved):

PR #1328, opened 2024-09-25
Add `Price Calculation` section.

zedzior:
Internal issue: https://linear.app/saleor/issue/SHOPX-918
This PR:
1. describes how Saleor calculates prices
2. extends overview discount page
3. adds `Manual discounts` section
4. adds paragraph `Tax data validation` (https://linear.app/saleor/issue/SHOPX-1348)
5. adds a section with complex examples

### DIFF
--- a/docs/developer/discounts/manual-discounts.mdx
+++ b/docs/developer/discounts/manual-discounts.mdx
@@ -5,14 +5,13 @@ sidebar_label: Manual Discounts
 
 ## Introduction
 Manual discounts can be applied by staff users to orders with `DRAFT` or `UNCONFIRMED` status.
-The discount can't be combined and overrides all the others discounts applied to the same object.
-The discount can be applied to the order, decreasing its subtotal and shipping price, or to the order line,
-reducing its unit price.
+These discounts cannot be combined with other discounts and override any existing ones.
+The discount can be applied to the order to decrease its subtotal and shipping price, or to the order line to reduce its unit price.
 
 ### Manual line discount
 The manual line discount can be applied with [`orderLineDiscountUpdate`](api-reference/orders/mutations/order-line-discount-update.mdx) mutation.
-It cuts unit price of the order line. The basis for the discount is undiscounted unit price, meaning any other unit price reductions like catalogue promotions or
-specific product vouchers are overridden.
+This discount reduces the unit price of the order line. The calculation is based on the undiscounted unit price, 
+which means that any other unit price reductions, such as catalogue promotions or specific product vouchers, will be overridden.
 
 #### Example
 
@@ -170,8 +169,8 @@ As a result, the discount reduced line 1 unit price from 50$ to 40$.
 
 ### Manual order discount
 The manual order discount can be applied with [`orderDiscountAdd`](api-reference/orders/mutations/order-discount-add.mdx) mutation.
-It affects both subtotal and shipping price. When the order qualify for order promotion or has entire order voucher applied, 
-the manual order discount takes precedence and override them. The basis for the calculation are [base subtotal](developer/price-calculation.mdx#discount-application) 
+It impacts both the subtotal and shipping price. When the order qualifies for an order promotion or has an entire order voucher applied, 
+the manual order discount takes precedence and overrides them. It's important to note that the discount is calculated based on [base subtotal](developer/price-calculation.mdx#discount-application) 
 and [base shipping](developer/price-calculation.mdx#discount-application) prices, meaning [line-level](developer/discounts/overview.mdx#level) discounts are not overriden. 
 
 
--- a/docs/developer/discounts/manual-discounts.mdx
+++ b/docs/developer/discounts/manual-discounts.mdx
@@ -7,12 +7,14 @@ sidebar_label: Manual Discounts
 Manual discounts can be applied by staff users to orders with `DRAFT` or `UNCONFIRMED` status.
 The discount can be applied to the order to decrease its subtotal and shipping price (manual order discount), 
 or to the order line to reduce its unit price (manual line discount). 
-These discounts cannot be combined with other discounts associated with the same object (order or line) and override any existing ones.
+These discounts are not combined with other discounts associated with the same object (order or line).
+Manual discount takes precedence and overrides any existing ones.
 
 ### Manual line discount
 The manual line discount can be applied with [`orderLineDiscountUpdate`](api-reference/orders/mutations/order-line-discount-update.mdx) mutation.
 This discount reduces the unit price of the order line. The calculation is based on the undiscounted unit price, 
 which means that any other unit price reductions, such as catalogue promotions or specific product vouchers, will be overridden.
+If the discount exceeds undiscounted unit price, it is set to 0 and don't have an impact on other lines.
 
 #### Example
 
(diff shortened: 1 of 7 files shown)


docs/developer/discounts/manual-discounts.mdx
```diff
@@ -0,0 +1,346 @@
...
+    }
+}
+```
+As a result, the discount reduced line 1 unit price from 50$ to 40$. 
+
+### Manual order discount
+The manual order discount can be applied with [`orderDiscountAdd`](api-reference/orders/mutations/order-discount-add.mdx) mutation.
+It impacts both the subtotal and shipping price. When the order qualifies for an order promotion or has an entire order voucher applied, 
+the manual order discount takes precedence and overrides them. It's important to note that the discount is calculated based on [base subtotal](developer/price-calculation.mdx#discount-application) 
+and [base shipping](developer/price-calculation.mdx#discount-application) prices, meaning [line-level](developer/discounts/overview.mdx#level) discounts are not overriden.
```
lkostrowski, line 174: so does it mean they are accumulated? Maybe we should write a paragraph how it works when both line and order discounts are applied

docs/developer/price-calculation.mdx
```diff
@@ -0,0 +1,96 @@
...
+The only thing Saleor calculates are undiscounted gross prices based on received tax rates.
+
+### Avatax plugin
+
+Avatax plugin is a built-in integration with Avalara. Similar to custom tax apps, final data relies on extrenal tax service,
+but in this case Saleor is responsible for handling Avalara response data and applying it to the prices.
+
+:::warning
+The Avatax plugin will be deprecated in Saleor 4.0.
+:::
```
lkostrowski, line 96: I don't think this is true. We need to confirm it, but I think it's deprecated *now* (app should be used) and it will be *removed* - probably before 4.0

docs/developer/extending/webhooks/synchronous-events/tax.mdx
```diff
@@ -443,3 +447,11 @@ The payload sent to the app would look like this:
...
    2. `total_gross_amount` - Gross price of the line.
    3. `total_net_amount` - Net price of the line.
+
+
+## Tax data validation
+
+The app response is considered valid if:
+1. There are no negative values
+2. Request's number of lines matches number of lines from response
+3. The prices and rates do not exceed database limits
```
korycins, line 457: Would be better to explicitly say: that prices should have maximal X numbers after coma etc. Right now it doesn't say anything as you need to be familiar with Saleor core code, to find the limits. 
Btw. Do we have such limitation? I though that we will get the price amount and apply rounding on it to reduce the decimal numbers 🤔

docs/developer/discounts/overview.mdx
```diff
@@ -6,9 +6,56 @@ sidebar_label: Overview
...
+    * [`specific product`](developer/discounts/vouchers.mdx#applying-the-specific-product-voucher)
+    * [`vouchers with applyOncePerOrder=True`](developer/discounts/vouchers.mdx#applying-the-once-per-order-entire-order-voucher)
+    * [`shipping`](developer/discounts/vouchers.mdx#applying-the-shipping-voucher)
+* [`manual discounts`](developer/discounts/manual-discounts.mdx)
+    * [`manual order discount`](developer/discounts/manual-discounts.mdx#manual-order-discount)
+    * [`manual line discount`](developer/discounts/manual-discounts.mdx#manual-line-discount)
+
+### Level
+We can distinguish discounts by the object they are associated with. 
+Line-level discounts are applicable first, directly to the order line, and decrease the line base prices.
```
korycins, line 39: `line base prices.`  It is not clear what is `base price`.

docs/developer/price-calculation.mdx
fowczarek, line 1: We still have inconsistency in naming. On line-level we have `base_unit_price = undiscounted_price - discounts` 
on order-level we have `untaxed_subtotal = base_subtotal - discounts`. Fixing inconsistency is part of those changes or not?

docs/developer/price-calculation.mdx
```diff
@@ -0,0 +1,96 @@
...
+
+### Untaxed price
+
+Having line-level discounts applied, we can sum up all the lines and calculate the base subtotal, which will be the basis for applying [`order-level discounts`](developer/discounts/overview.mdx#level):
+
+`untaxed_subtotal` = `base_subtotal` - (`entire_order_voucher` or `order_promotion` or `manual_discount`)
+
+`untaxed_shipping` = `base_shipping_price` - `manual_discount`
+
+All order-level discounts are mutually exclusive and can't be combined.
```
fowczarek, line 60: IMO this section should be a Warning block.

docs/developer/price-calculation.mdx
```diff
@@ -0,0 +1,96 @@
...
+
+* The entire order voucher decreases the subtotal price and doesn't affect the shipping price.
+* The order promotion can either decrease the subtotal or grant a gift, which will be added to the order as an extra order line with zero price.
+* The manual discount decreases both subtotal and shipping prices. If the fixed value is provided, Saleor proportionally computes the subtotal and shipping portion.
```
fowczarek, line 64: Should we also explain this proportion on `Manual Discounts` page and link it here?

docs/developer/price-calculation.mdx
fowczarek, line 96: Maybe?
```suggestion
:::warning
The Avatax plugin will be deprecated in Saleor 4.0.
:::

Avatax plugin is a built-in integration with Avalara. Similar to custom tax apps, final data relies on extrenal tax service,
but in this case Saleor is responsible for handling Avalara response data and applying it to the prices.
```

docs/developer/discounts/manual-discounts.mdx
```diff
@@ -5,8 +5,9 @@ sidebar_label: Manual Discounts
 
 ## Introduction
 Manual discounts can be applied by staff users to orders with `DRAFT` or `UNCONFIRMED` status.
-These discounts cannot be combined with other discounts and override any existing ones.
-The discount can be applied to the order to decrease its subtotal and shipping price, or to the order line to reduce its unit price.
+The discount can be applied to the order to decrease its subtotal and shipping price (manual order discount), 
+or to the order line to reduce its unit price (manual line discount). 
+These discounts cannot be combined with other discounts associated with the same object (order or line) and override any existing ones.
```
IKarbowiak, line 10: It sounds like I cannot apply a manual discount on an order with a voucher. Maybe something like this will be better:
```suggestion
These discounts are not combined with other discounts associated with the same object (order or line), manual discount takes precedence and overrides any existing ones.
```

docs/developer/discounts/manual-discounts.mdx
```diff
@@ -0,0 +1,348 @@
...
+Request:
+
+```graphql
+mutation OrderLineDiscountUpdate {
+  orderLineDiscountUpdate(
+    orderLineId: "T3JkZXJMaW5lOmFlYTBmNjMzLWU0MTItNDkwZS1hODhhLThlMTg3YzQ3N2ZkNQ==", 
+    input: {
+        valueType: PERCENTAGE,
+        value: 20,
+        reason: "staff line discount"
```
zedzior, line 35: Move to variables

docs/developer/discounts/manual-discounts.mdx
```diff
@@ -0,0 +1,346 @@
...
+                            "amount": 0.0
+                        }
+                    }
+                ]
+            }
+        }
+    }
+}
+```
+As a result, the discount reduced line 1 unit price from 50$ to 40$.
```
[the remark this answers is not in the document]
zedzior: Let's create separate section like Discount Combination or complex examples or sth

zedzior:
Made it draft as I have to add extra section with complex examples

lkostrowski (approved):

aniav (approved):
A lot of good content 👌

szczecha (approved):
❤️ 🥹

PR #1330, opened 2024-10-01
Improve ProductVariant.privateMetadata permission info

Cloud11PL:


kadewu (commented):
docs/api-reference/ is automatically generated.
We need to adjust it in the saleor core repository.

And probably actually fix the issue with genereated `permissions` on core side as all metadata fields are affected by it, and only some of them gives false-positive result on that.

PR #1355, opened 2024-10-11
Docusaurus faster

krzysztofwolski:
- PR uses bleeding edge ducusaurus build
- more context why its faster: `https://github.com/facebook/docusaurus/issues/10556`
- had to fight with dependencies, thats why I've added `legacy-peer-deps=true`


krzysztofwolski:
Comparison of a few latest builds - this PR is on top:
image

PR #1359, opened 2024-10-15
Rework app examples

Droniu:
This PR improves App Examples page with GitHub-style cards.
image
image


peelar:
Will there be a working deployment?

peelar (changes requested):
I like how these cards look in general but I have three issues with them:

1. IMO, the teal color of the technology tag is a bit out of character for our docs design. I would keep it more monochromatic (a different shade of grey, for example).
2. We already have a `<CardGrid>` component that is universally used across several pages in the docs. Now, we introduce new cards that look significantly different from those. Your cards definitely look better but maybe they should extend the basic ones, so we have some similarity.
3. **(blocking)** When you hover the card, the animation tilts the layout of the page.

Droniu:
- Teal has been replaced with gray
- Layout shift issue fixed (although I had to use `!important`)
- Card bg color now changes on hover, adding more contrast and interactivity
- In terms of CardGrid component, I think the new one should be used to represent GitHub repositories only, so their use cases are different. We can discuss it further

docs/developer/extending/apps/developing-apps/self-hosting-saleor-apps.mdx
```diff
@@ -0,0 +1,117 @@
+---
+title: Self-hosting Saleor Apps
+---
+
+# Self-hosting Saleor Apps
```
lkostrowski, line 5: some of the apps may have more advanced setup. I think its good to have an introduction article but eventually we need to redirect to the specific apps for details. Eg some apps will require dynamodb

docs/developer/extending/apps/developing-apps/self-hosting-saleor-apps.mdx
```diff
@@ -0,0 +1,117 @@
...
+
+{/* TODO add `apps` repository with setup */}
+
+You can find the code for all the apps from the Saleor App Store in the [`saleor/apps` monorepo](https://github.com/saleor/apps).
+
+Each app is a [Next.js](https://nextjs.org/docs/getting-started) application based off [saleor-app-template](https://github.com/saleor/saleor-app-template).
+
+In this article, you will learn how to run the Saleor App Store apps locally.
```
lkostrowski, line 13: page claims to be about "self hosting" while this is about running locally

lkostrowski (dismissed):

docs/developer/extending/apps/developing-apps/self-hosting-saleor-apps.mdx
Droniu: That was an existing guide that I moved to separate section, so that we have seperate examples and guide how to run app store apps locally

peelar (approved):

PR #1379, opened 2024-11-04
Update payment lifecycle docs

witoszekdev:
Our current payment lifecycle docs are incorrect (e.g. using "refund" when talking about "cancel" behavior), this PR fixes the description to match Saleor's current behavior.

It also adds better description for `chargeStatus` and `authorizationStatus` fields on both Checkout and Order - and explains differences between each.

It also adds better descriptions for Saleor's metal model and assumptions about Transaction API: about flexibility vs strict validation.


docs/developer/payments/lifecycle.mdx
```diff
@@ -173,27 +258,126 @@ it may still remain unpaid if the pending amounts fail to process.
...
+- Reporting the same event based on `type` and `pspReference` with different `amount`
+  - Saleor will throw an exception: "The transaction with provided `pspReference` and `type` already exists with different amount."
+
+:::note
+**Example**: `REFUND_SUCCESS` with `pspReference: psp1` and `amount: 10 USD`, then report `REFUND_SUCCESS` with `pspReference: psp1` and `amount: 21 USD`
+:::
+
+:::tip
+**Fix**: Previous event must be invalidated via `_FAILURE` event, then report `_SUCCESS` event again
+:::
```
witoszekdev, line 379: I'm not sure about this if this is in fact a fix - what should we do when we want to update event's amount with the same `pspReference`? 🤔

docs/developer/payments/lifecycle.mdx
```diff
@@ -173,27 +258,126 @@ it may still remain unpaid if the pending amounts fail to process.
...
+
+#### Automatic events
+
+Events are created automatically by Saleor when calling `transactionInitialize` mutation. It creates either:
+- `AUTHORIZATION_REQUEST` when `action` in mutation input or `channel.transactionFlowStrategy` is `AUTHORIZATION`
+- or `CHARGE_REQUEST` when `action` in mutation input or `channel.transactionFlowStrategy` is `CHARGE`
+
+Events are also created when requesting action in Saleor (refund, charge, cancel) via [`transactionRequestAction`](/api-reference/payments/mutations/transaction-request-action.mdx) mutation.
+
+Such events don't have initially any `pspReference`, it's filled in once Saleor receives a response from payment app.
```
witoszekdev, line 290: This can be problematic, in case app returns `CHARGE_ACTION_REQUIRED`, then `CHARGE_REQUEST` with `pspReference`. Saleor will update the `CHARGE_REQUEST` event with returned `pspReference`, but now it will be _before_ the `CHARGE_ACTION_REQUIRED`, which is incorrect.

We have a ticket to look into this: https://linear.app/saleor/issue/SHOPX-1580/transactionprocess-shouldnt-mess-with-transactionitem-event-history

Should we add a note about this in our docs? 🤔

docs/developer/payments/lifecycle.mdx
```diff
@@ -173,27 +258,126 @@ it may still remain unpaid if the pending amounts fail to process.
...
+
+- `authorizedAmount` - `AUTHORIZATION_SUCCESS` event amount
+  - there can be only one `AUTHORIZATION_SUCCESS` event for given TransactionItem
+  - it can be modified by `AUTHORIZATION_ADJUSTMENT` event
+  - subtracting `canceledAmount`
+- `authorizePendingAmount` - sum of `AUTHORIZATION_REQUEST` event amounts
+- `chargedAmount` - sum of `CHARGE_SUCCESS` event amounts
+  - and sum `REFUND_REVERSE` event amounts
+  - subtracting all `CHARGE_BACK` events that lower this amount
+  - subtracting `refundedAmount`
```
korycins, line 323: IMHO it is not clear to me. It is not like that `refundedAmount` has impact on `chargedAmount`. It is more like:
- Having REFUND_SUCCESS event would subtract the `event.amount` from the calculated `chargedAmount` and add `event.amount` to refundedAmount. 
- Having pending REFUND_REQUEST event would subtract the `event.amount` from the calculated `chargedAmount` add `event.amount` to refundPendingAmount.

docs/developer/payments/lifecycle.mdx
```diff
@@ -173,27 +258,126 @@ it may still remain unpaid if the pending amounts fail to process.
...
+
+:::note
+`AUTHORIZATION_ACTION_REQUIRED`, `CHARGE_ACTION_REQUIRED`, and `INFO` event types are not included in amount calculations.
+:::
+
+To learn more about each event type, see [Transaction event types](#transaction-event-types) section.
+
+#### Events deduplication
+
+Each event is identified by `pspReference`. This field is used to de-duplicate events in case there were reported multiple times by the payment app.
```
korycins, line 339: Maybe worth mentioning that the deduplication works across single `TransactionItem`.

docs/developer/payments/lifecycle.mdx
```diff
@@ -173,27 +258,126 @@ it may still remain unpaid if the pending amounts fail to process.
...
+
+For example if Saleor receives two `CHARGE_SUCCESS` events with the same `pspReference` and `amount` there will be only one event stored on Transaction. Second event will be treated as "already reported" and it won't be stored by Saleor.
```
korycins, line 341: It is worth mentioning, that providing two events with CHARGE_SUCCESS, same `pspReference` but with different `amount` will cause an error.

docs/developer/payments/lifecycle.mdx
korycins: We do not have EDIT option. Events are treated like history. If you want to modify the value, then you would need to work with existing event types. In your case, you will need to provide different pspReference i guess.

witoszekdev: OK, so the new `SUCCESS` event **must have** other `pspReference` otherwise it will be ignored by Saleor? @korycins

witoszekdev: I've just quickly checked this and nope, as you said it must have a different `pspReference` :/

```
Error returned when reporting event:
The transaction with provided `pspReference` and `type` already exists with different amount.
```

docs/developer/payments/lifecycle.mdx
```diff
@@ -56,107 +56,192 @@ Creating the refund will change the status to:
...
-For a checkout, authorization means that funds have been either authorized or charged (or a combination of partially charged and partially authorized).
-The status takes into account the following fields: `authorizedAmount`, `chargedAmount`, `authorizePendingAmount`, and `chargePendingAmount`:
+Based on that calculation `Checkout.authorizeStatus` can have the following values:
 
-- **NONE**: No funds have been authorized.
-- **PARTIAL**: The available funds do not fully cover the checkout's total amount.
-- **FULL**: The funds fully cover the checkout's total amount. At this point, `checkoutComplete` can be called to convert the checkout into an order. 
+- **`NONE`**: No funds have been authorized. (`CoveredAmount <= 0`)
+- **`PARTIAL`**: Some funds were authorized, but they do not cover required amount (`AmountToCover > CoveredAmount`)
+- **`FULL`**: The authorized funds fully cover the required amount. (`AmountToCover == CoveredAmount`)
```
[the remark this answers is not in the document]
witoszekdev: Fixed comparison 👍🏻

docs/developer/payments/lifecycle.mdx
```diff
@@ -173,27 +258,126 @@ it may still remain unpaid if the pending amounts fail to process.
...
+- `AUTHORIZATION_REQUEST` when `action` in mutation input or `channel.transactionFlowStrategy` is `AUTHORIZATION`
+- or `CHARGE_REQUEST` when `action` in mutation input or `channel.transactionFlowStrategy` is `CHARGE`
+
+Events are also created when requesting action in Saleor (refund, charge, cancel) via [`transactionRequestAction`](/api-reference/payments/mutations/transaction-request-action.mdx) mutation.
+
+Such events don't have initially any `pspReference`, it's filled in once Saleor receives a response from payment app.
+
+Saleor also creates automatically `_FAILURE` events when an exception happens:
+- when handling `transactionEventReport` mutation
+- when handling incorrect response from [sync webhook sent to payment app](/developer/extending/webhooks/synchronous-events/transaction.mdx)
```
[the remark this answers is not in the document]
witoszekdev: I can check this 👀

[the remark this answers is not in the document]
witoszekdev: I've updated the description, listing what we mean by "failure"

docs/developer/payments/lifecycle.mdx
```diff
@@ -173,27 +258,126 @@ it may still remain unpaid if the pending amounts fail to process.
...
+
+Such events don't have `pspReference` and include a `message` that explains the error.
+
+#### App events
+
+Events are also created by a payment app that processes the payment:
+
+- When app sends response to any of the [Transaction synchronous events](/developer/extending/webhooks/synchronous-events/transaction.mdx)
+  - Saleor will either fill missing `pspReference` on the automatically created event that already exists
```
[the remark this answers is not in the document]
witoszekdev: I've updated the description to be way more verbose about this

docs/developer/payments/lifecycle.mdx
```diff
@@ -173,27 +258,126 @@ it may still remain unpaid if the pending amounts fail to process.
...
+  - or will create a new event if it has different `type` (e.g. user requested `AUTHORIZATION`, but app returned `CHARGE_SUCCESS` - we'll have two events `AUTHORIZATION_REQUEST` with no `pspReference` and `CHARGE_SUCCESS`)
+- When [`transactionEventReport` mutation is called](/developer/payments/transactions.mdx#reporting-actions-for-transactions) (usually happens when payment provider sends asynchronous webhook)
+
+### Rules of calculating amounts
+
+Each event impacts different amount on `TransactionItem`:
+
+:::caution
+Each event can be excluded from calculation if it has matching `_FAILURE` event with the same `pspReference` and newer `time`.
```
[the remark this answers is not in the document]
witoszekdev: Fixed 👍🏻

docs/developer/payments/lifecycle.mdx
witoszekdev: hmm, ok, could we word this differently then? 🤔 

I think in context of `chargedAmount` it's actually easier to subtract `refundedAmount` and `refundPendingAmount` rather than: "if you have REFUND_SUCCESS, then it subtracts event's value from chargedAmount and puts it in refundedAmount".

Correct me if I'm wrong but this is also how this calculation works in Saleor's code, right? We don't operate on individual events but on sum (`refundedAmount`, `canceledAmount`, etc.)

docs/developer/payments/lifecycle.mdx
witoszekdev: Fixed 👍🏻

docs/developer/payments/lifecycle.mdx
witoszekdev: Nevermind, I've adedd an example that explains just that.

docs/developer/payments/lifecycle.mdx
korycins: > Correct me if I'm wrong but this is also how this calculation works in Saleor's code, right? We don't operate on individual events but on sum (refundedAmount, canceledAmount, etc.)

We calculate the values based on the events that we have. We matched them by PSP reference and their type, and apply different action for amounts. 
The amounts like `chargedAmount`, `refundedAmount`, `chargePendingAmount` etc are the result of calculations created based on the events. 


> I think in context of chargedAmount it's actually easier to subtract refundedAmount and refundPendingAmount rather than: "if you have REFUND_SUCCESS, then it subtracts event's value from chargedAmount and puts it in refundedAmount".

If we want to keep this simply, then I guess it is okay, but I would suggest provide some different structure. I mean, having entry like: `- subtracting refundedAmount` , wasn't very clear for me. Personally I would expect to have info in section for `refundedAmount` that would say, that it reduces the value of `chargedAmount`, same for `refundPendingAmount`.

docs/developer/payments/lifecycle.mdx
korycins: But this is expected. You reported the charge action with PSP:123. This is the ID of this action. As you received it from payment-provider. Dont see a situation where payment-provider returns you a different value 🤔  At least this is what we expect when we were building the interface. 
Above you explain that events are like bank account history, which is really good example, but now you expect, that we will give you an option to modify the "history" 😅

korycins (dismissed):

docs/developer/payments/lifecycle.mdx
witoszekdev: Sure, I don't want to drag this any longer, we can do this in separate PR.

I personally find having global description talking about amounts easier to understand than reading entire events description, but if this is incorrect we should point out that Saleor itself doesn't operate on these amounts, but rather adds or subtracks from TransactionItem amounts while iterating over events? I don't know how to put it in words to be honest 😅 

btw. Our `REFUND_SUCCESS` section already describes this:

> If there is no related `REFUND_REQUEST` event, the provided `amount` will reduce `transaction.chargedAmount`.

Is this this right though? I think we decrease `chargedAmount` anyway? Or does it mean we decrease `chargedAmount` first on `REFUND_REQUEST`, so when there's matching `REFUND_SUCCESS` there's nothing to be changed?

docs/developer/payments/lifecycle.mdx
```diff
@@ -56,107 +56,193 @@ Creating the refund will change the status to:
...
-Authorization means that funds have been either fully or partially authorized or charged.
-It uses the fields `authorizedAmount` and `chargedAmount`, and the calculation is adjusted by subtracting `order.totalGrantedRefund` from the total.
+- sum of **charged** transactions (`TransactionItem.chargedAmount`)
+- sum of **authorized** transactions (`TransactionItem.authorizedAmount`)
+- `order.totalGrantedRefund` value
+- `order.total` value
 
 :::caution
-Unlike checkout, this status does not consider pending amounts.
+Unlike Checkout, this status does not take into consideration pending transactions.
```
IKarbowiak, line 133: Maybe we should mention here that it might result in situation that order completed from a  checkout with `authorizationStatus: FULL` will have `authorizationStatus: NONE`

IKarbowiak (dismissed):

docs/developer/payments/lifecycle.mdx
```diff
@@ -56,107 +56,193 @@ Creating the refund will change the status to:
...
-The status is calculated based on all transactions assigned to the checkout.
-It takes into account the fields: `chargedAmount` and `chargePendingAmount`.
+It uses a formula:
+
+```
+AmountToCover = order.total - order.totalGrantedRefund
+CoveredAmount = sum(order.transactions.chargedAmount)
+```
+
+Based on that calculation `Order.cahrgeStatus` can have following values:
```
krzysztofzuraw, line 206:
```suggestion
Based on that calculation `Order.chargeStatus` can have following values:
```

docs/developer/payments/lifecycle.mdx
```diff
@@ -173,27 +259,152 @@ it may still remain unpaid if the pending amounts fail to process.
...
+
+:::note
+It can look like this:
+1. User requests payment by calling `transactionInitialize` mutation - Saleor immediately created new `CHARGE_REQUESTED` event
+2. App sends response from `TRANSACTION_INITIALIZE_SESSION` webhook with `type`: `CHARGE_ACTION_REQUIRED`
+3. User completes the action and calls `transactionProcess`
+4. App sends response from `TRANSACTION_PROCESS_SESSION` webhook with the same type as first event - `CHARGE_REQUESTED`
+5. Saleor instead of creating new event, updates existing event `CHARGE_REQUESTED` with received `pspReference`
+
+Note that this will result in events order:
```
krzysztofzuraw, line 321: **Question:** you mean here order of events on checkout / order? If so maybe 

`Note that this will result in events ordered as follows`?

docs/developer/payments/lifecycle.mdx
```diff
@@ -173,27 +259,152 @@ it may still remain unpaid if the pending amounts fail to process.
...
+With that in mind, Saleor prohibits following changes, that would otherwise cause incorrect internal calculations:
+- Multiple `AUTHORIZATION_SUCCESS` events with different either `pspReference` or `amount`
+  - Saleor will throw an exception: "Event with `AUTHORIZATION_SUCCESS` already reported for the transaction. Use `AUTHORIZATION_ADJUSTMENT` to change the authorization amount."
+
+:::note
+**Example**: `AUTHORIZATION_SUCCESS` event with `pspReference: psp1` and `amount: 10 USD`, then report `AUTHORIZATION_SUCCESS` event with `pspReference: psp2` and `amount: 10 USD`
+:::
+
+:::tip
+**Fix**: In order to change `authorizedAmount` value, payment app should send `AUTHORIZATION_ADJUSTMENT` event. If there are in fact two valid `AUTHORIZATION_SUCCESS` events, create new `TransactionItem`
```
krzysztofzuraw, line 394: should app create new `TransactionItem` or Saleor will do it automatically?

docs/developer/payments/lifecycle.mdx
witoszekdev: Yes, I know we have a history ledger and it's not modifiable - meaning that you cannot **edit** existing events, **but** we probably should have an option to make adjustments, that allow changing this history **by adding new events**.

Going back to the bank analogy: if they have some issue on their side, and you make a claim (e.g. they charged you more for some service than they should like ATM payout) they can make an correction for that charge by giving you money back. Same thing with invoices, if you made a mistake when issuing an invoice, you issue a correction detailing the changes.

witoszekdev: It's out of scope for this PR, here's a link for this feature idea: https://linear.app/saleor/issue/IFREQ-36/corrections-for-transactionevents

docs/developer/payments/lifecycle.mdx
witoszekdev: Yup, I give an example situation and what what will be the order of `TransactionEvent`s in Saleor.

docs/developer/payments/lifecycle.mdx
witoszekdev: Nope, this has to be done by app - Saleor will reject `transactionEventReport` mutation if app would report 2 `AUTHORIZATION_SUCCESS` events

IKarbowiak (approved):

PR #1380, opened 2024-11-08
Add `Resolving Paid Checkouts Not Converting to Orders` section to checkout troubleshooting

IKarbowiak:
Describe solutions for paid checkout not converted into an order.

### DIFF
--- a/docs/developer/checkout/troubleshooting.mdx
+++ b/docs/developer/checkout/troubleshooting.mdx
@@ -11,3 +11,52 @@ sidebar_label: Troubleshooting
 ### Variant/Product is not available
 
 Make sure product is available in the channel the checkout is created. See [Product Troubleshooting](/developer/products/troubleshooting.mdx).
+
+### Resolving Paid Checkouts Not Converting to Orders
+
+Sometimes, a `Checkout` that has been paid might not convert into an `Order`.
+This can happen for various reasons, such as the customer closing the browser before the `checkoutComplete` mutation was called.
+To ensure that fully paid checkouts are reliably converted into orders, consider implementing one of the following solutions:
+
+#### Solution 1: Use a Webhook to Complete Checkouts
+
+Set up the `CHECKOUT_FULLY_PAID` webhook, and create an app to call the `checkoutComplete` mutation when this webhook is triggered.
+	•	For more information on setting up webhooks, refer to the [webhook overview](/developer/extending/webhooks/overview).
+	•	You can find guidance on app development [here](/developer/extending/apps/overview).
+
+	Note: This approach only works when the `CHARGE` is used as the default transaction flow strategy.
+
+#### Solution 2: Enable Automatic Checkout Completion
+
+Enable [automatic checkout completion](/developer/payments/transactions#automatic-checkout-completion)
+to automatically convert a `Checkout` into an `Order` once it is fully paid.
+
+A checkout is considered fully paid when `Checkout.authorizeStatus` is set to `FULL`, meaning the associated `TransactionItems` cover the `checkout.totalPrice`.
+
+#### Solution 3: Periodically Query Uncompleted Fully Paid Checkouts
+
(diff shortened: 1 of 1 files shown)


docs/developer/checkout/troubleshooting.mdx
```diff
@@ -11,3 +11,52 @@ sidebar_label: Troubleshooting
...
+	•	You can find guidance on app development [here](/developer/extending/apps/overview).
+
+	Note: This approach only works when the `CHARGE` is used as the default transaction flow strategy.
+
+#### Solution 2: Enable Automatic Checkout Completion
+
+Enable [automatic checkout completion](/developer/payments/transactions#automatic-checkout-completion)
+to automatically convert a `Checkout` into an `Order` once it is fully paid.
+
+A checkout is considered fully paid when `Checkout.authorizeStatus` is set to `FULL`, meaning the associated `TransactionItems` cover the `checkout.totalPrice`.
```
lkostrowski, line 34: is that true for both transaction flows (authorize and charge)?

lkostrowski (approved):

docs/developer/checkout/troubleshooting.mdx
IKarbowiak: Yes. When `authorizeStatus` is updated, both `total_authorized` and `total_charged` are taken into consideration.

docs/developer/checkout/troubleshooting.mdx
```diff
@@ -11,3 +11,52 @@ sidebar_label: Troubleshooting
...
+This can happen for various reasons, such as the customer closing the browser before the `checkoutComplete` mutation was called.
+To ensure that fully paid checkouts are reliably converted into orders, consider implementing one of the following solutions:
+
+#### Solution 1: Use a Webhook to Complete Checkouts
+
+Set up the `CHECKOUT_FULLY_PAID` webhook, and create an app to call the `checkoutComplete` mutation when this webhook is triggered.
+	•	For more information on setting up webhooks, refer to the [webhook overview](/developer/extending/webhooks/overview).
+	•	You can find guidance on app development [here](/developer/extending/apps/overview).
+
+	Note: This approach only works when the `CHARGE` is used as the default transaction flow strategy.
```
witoszekdev, line 27: I'm not sure we should recommend this approach, since queue might be overloaded - this has previously resulted in situations where webhooks weren't delivered for 6h+ and because of that our release funds task that refunds transactions for uncompleted checkouts.

This also contradicts our previous recommendations, for example [in migration guide](https://docs.saleor.io/upgrade-guides/core/3-19-to-3-20#automatic-checkout-completion):
> Previously a store might have had missing orders, due to checkoutComplete mutation not being always called. This could have happened for many unexpected reasons: network issues, storefront bug, customer leaving website, closing tab before being redirected from payment provider, etc.

docs/developer/checkout/troubleshooting.mdx
```diff
@@ -11,3 +11,52 @@ sidebar_label: Troubleshooting
...
+
+#### Solution 2: Enable Automatic Checkout Completion
+
+Enable [automatic checkout completion](/developer/payments/transactions#automatic-checkout-completion)
+to automatically convert a `Checkout` into an `Order` once it is fully paid.
+
+A checkout is considered fully paid when `Checkout.authorizeStatus` is set to `FULL`, meaning the associated `TransactionItems` cover the `checkout.totalPrice`.
+
```
witoszekdev, line 35: suggestion: IMHO this should be our **default** recommendation

witoszekdev (changes requested):

docs/developer/checkout/troubleshooting.mdx
IKarbowiak: So maybe we should not delete this recommendation, but move it to the end and mention the limitations?

lkostrowski (approved):

witoszekdev (approved):

PR #1381, opened 2024-11-08
Add TransactionItem and  `TransactionEvents` `pspReference` Matching section on `Transactions` page

IKarbowiak:
Describe how we match `pspRefereneces`

### DIFF
--- a/docs/developer/payments/transactions.mdx
+++ b/docs/developer/payments/transactions.mdx
@@ -450,6 +450,77 @@ Once `Checkout` is converted to into `Order`, the `authorizeStatus` might change
 - **For orders**: pending Transactions **are not** considered "paid" and will not be included in the [`Order.authorizeStatus`](/api-reference/orders/enums/order-authorize-status-enum.mdx#values) calculation.
 :::
 
+### TransactionItem and  TransactionEvents pspReference Matching
+
+Both [TransactionItem](/api-reference/payments/objects/transaction-item)
+and [TransactionEvent](/api-reference/payments/objects/transaction-event)
+contain a `pspReference` field, which refers to the payment service provider reference.
+	-	The `pspReference` on `TransactionItem` identifies the overall payment.
+	-	The `pspReference` on `TransactionEvent` identifies specific actions that occurred within the `Transaction`.
+
+**Key Assumptions:**
+- Only one `TransactionItem` instance can exist with a given `pspReference`.
+- For a specific `TransactionItem`, only one `TransactionEvent` of a given `type` and `pspReference` may exist, except for the following types:
+  -	`AUTHORIZATION_ACTION_REQUIRED`
+	-	`CHARGE_ACTION_REQUIRED`
+	-	`INFO`
+- Only one `TransactionEvent` of type `AUTHORIZATION_SUCCESS` is allowed to exist.
+
+#### Setting the `pspReference` on the `TransactionItem`
+
+The pspReference value on a `TransactionItem` can initially be `None`.
+When the first `TransactionEvent` containing a `pspReference` is reported,
+this value is automatically set on the associated `TransactionItem`.
+
+The `pspReference` can be updated through the `transactionUpdate` mutation if necessary.
+
+#### Events with optional `pspReference` 
+
+The following events do not require a `pspReference`:
+- `CHARGE_ACTION_REQUIRED`
+- `AUTHORIZATION_ACTION_REQUIRED`
+- `CHARGE_FAILURE`
+- `AUTHORIZATION_FAILURE`
+- `REFUND_FAILURE`
+- `CANCEL_FAILURE`
+
+If `pspReference` is missing, the event is not considered in transaction amount recalculations.
+
+#### Handling TransactionEvent Reports for Existing Types
+
+When a `TransactionEvent` is reported:
+	1.	**Event Type Check:** If the event type is `AUTHORIZATION_ACTION_REQUIRED`, `CHARGE_ACTION_REQUIRED`, or `INFO`, a new event is always created.
+	2.	**Existing Event Check:** For other types, the system searches for an existing event with the same `type` and `pspReference`.
+	    -	**Amount Match:** If an existing event is found, the system compares its amount to the reported event’s amount.
+	        - If the amounts match, the existing event is reused.
+	        -	If they do not match, an error is raised.
+	    -	**No Existing Event:** If no matching event is found, a new one is created, except for `AUTHORIZATION_SUCCESS`, which must be unique, attempting to create a duplicate raises an error.
+
+```mermaid
+flowchart TD
+    A[TransactionEvent report] --> B{
+      The event type is
+      AUTHORIZATION_ACTION_REQUIRED,
+      CHARGE_ACTION_REQUIRED, or INFO}
+    B -->|YES| D[Create a new event]
+    B -->|NO| E{
+      The event with provided
+      pspReference and type
+      already exist}
+    E -->|YES| F{
+      The provided amount
+      match the existing
+      event amount}
+    E -->|NO| G{
+      The event type
+      is AUTHORIZE_SUCCESS
+      }
+    G --> |NO| D
+    G --> |YES| H[Raise an error]
+    F --> |YES| I[Return an existing event]
+    F --> |NO|H
+```
+
 ## Related resources
 
 - [Building Payment Apps](/developer/extending/apps/building-payment-app.mdx)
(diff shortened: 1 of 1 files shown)


docs/developer/payments/transactions.mdx
```diff
@@ -450,6 +450,77 @@ Once `Checkout` is converted to into `Order`, the `authorizeStatus` might change
...
+
+The following events do not require a `pspReference`:
+- `CHARGE_ACTION_REQUIRED`
+- `AUTHORIZATION_ACTION_REQUIRED`
+- `CHARGE_FAILURE`
+- `AUTHORIZATION_FAILURE`
+- `REFUND_FAILURE`
+- `CANCEL_FAILURE`
+
+If `pspReference` is missing, the event is not considered in transaction amount recalculations.
```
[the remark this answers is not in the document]
IKarbowiak: I can extend it a little, but I do not want to describe here how the recalculations are woking because it should be separate section just linked here.

docs/developer/payments/transactions.mdx
```diff
@@ -450,6 +450,80 @@ Once `Checkout` is converted to into `Order`, the `authorizeStatus` might change
...
+### TransactionItem and  TransactionEvents pspReference Matching
+
+Both [TransactionItem](/api-reference/payments/objects/transaction-item)
+and [TransactionEvent](/api-reference/payments/objects/transaction-event)
+contain a `pspReference` field, which refers to the payment service provider reference.
+	-	The `pspReference` on `TransactionItem` identifies the overall payment.
+	-	The `pspReference` on `TransactionEvent` identifies specific actions that occurred within the `Transaction`.
+
+**Key Assumptions:**
+- Only one `TransactionItem` instance can exist with a given `pspReference`.
```
korycins, line 462: Are you sure about this? I don't think that we have such condition. Maybe we have a validation where psp_reference+app_identifier, but this needs to be checked first.

docs/developer/payments/transactions.mdx
```diff
@@ -450,6 +450,80 @@ Once `Checkout` is converted to into `Order`, the `authorizeStatus` might change
...
+- For a specific `TransactionItem`, only one `TransactionEvent` of a given `type` and `pspReference` may exist, except for the following types:
+  -	`AUTHORIZATION_ACTION_REQUIRED`
```
korycins, line 464: Incorrect formatiing? 🤔

docs/developer/payments/transactions.mdx
```diff
@@ -450,6 +450,80 @@ Once `Checkout` is converted to into `Order`, the `authorizeStatus` might change
...
+	-	`CHARGE_ACTION_REQUIRED`
```
korycins, line 465: Maybe worth to extend the: https://docs.saleor.io/developer/payments/lifecycle#transaction-event-types with `AUTHORIZATION_ACTION_REQUIRED` and `CHARGE_ACTION_REQUIRED`. With similar info that we have for these events in https://docs.saleor.io/developer/payments/payment-apps#initialize-transaction. Then here we could  include the link to the event description? 🤔

docs/developer/payments/transactions.mdx
```diff
@@ -450,6 +450,80 @@ Once `Checkout` is converted to into `Order`, the `authorizeStatus` might change
...
+	-	`INFO`
+- Only one `TransactionEvent` of type `AUTHORIZATION_SUCCESS` is allowed to exist.
```
korycins, line 467: Would be also good to mention that in case of modifying the authorization, need to use https://docs.saleor.io/developer/payments/lifecycle#authorization_adjustment

docs/developer/payments/transactions.mdx
```diff
@@ -450,6 +450,80 @@ Once `Checkout` is converted to into `Order`, the `authorizeStatus` might change
...
+
+#### Setting the `pspReference` on the `TransactionItem`
+
+The pspReference value in a TransactionItem may initially be empty if it is not provided
+during the TransactionCreate mutation or as a result of the TransactionInitialize mutation.
+When the first `TransactionEvent` containing a `pspReference` is reported,
```
korycins, line 473: Is it also true for `transactionCreate`? 
Also maybe we could change the sentence, to say that transactionItem.psp_reference will be set based on first success event received from payment-app  (if this is true).

> When the first `TransactionEvent` containing a `pspReference` is reported,

Also this needs to be checked. I mean, we need to confirm if we always set first pspReference, or first success pspReference

docs/developer/payments/transactions.mdx
```diff
@@ -450,6 +450,80 @@ Once `Checkout` is converted to into `Order`, the `authorizeStatus` might change
...
+this value is automatically set on the associated `TransactionItem`.
+
+The `pspReference` can be updated through the `transactionUpdate` mutation if necessary.
+
+#### Events with optional `pspReference`
```
korycins, line 478: We have a section: https://docs.saleor.io/developer/payments/transactions#amount-calculations-for-reporting-an-action-with-missing-amount-value were we describe when amount can be optional. Isn't like this should also be there? This section described `Reporting actions for transactions`. Right now some info about optional fields is there, and some info is here 🤔 

Also, what about the "webhook-part": https://docs.saleor.io/developer/extending/webhooks/synchronous-events/transaction#response-4
Don't we need to also say there that the reference can be optional for specific events?

docs/developer/payments/transactions.mdx
```diff
@@ -450,6 +450,80 @@ Once `Checkout` is converted to into `Order`, the `authorizeStatus` might change
...
+The following events do not require a `pspReference`:
+- `CHARGE_ACTION_REQUIRED`
+- `AUTHORIZATION_ACTION_REQUIRED`
+- `CHARGE_FAILURE`
+- `AUTHORIZATION_FAILURE`
+- `REFUND_FAILURE`
+- `CANCEL_FAILURE`
+
+If `pspReference` is missing, the event will be created as usual but will be excluded
+from transaction amount recalculations. As a result, it will not affect
```
korycins, line 489: Maybe worth to mentioned, that missing PSPreference will not have an impact on existing transactionEvent with pspReference:
I mean the part that is described here:
image
Ref: https://docs.saleor.io/developer/payments/lifecycle#authorization-events

docs/developer/payments/transactions.mdx
```diff
@@ -450,6 +450,80 @@ Once `Checkout` is converted to into `Order`, the `authorizeStatus` might change
...
+`charge`, `refund`, `cancel`, or `authorize` amounts.
+
+#### Handling TransactionEvent Reports for Existing Types
```
korycins, line 492: The chapter name suggest me that you're describe some part of the transaction lifecycle. If you agree about that, then shouldn't we move this to: https://docs.saleor.io/developer/payments/lifecycle

docs/developer/payments/transactions.mdx
```diff
@@ -450,6 +450,80 @@ Once `Checkout` is converted to into `Order`, the `authorizeStatus` might change
...
+
+When a `TransactionEvent` is reported:
```
korycins, line 494: For https://docs.saleor.io/developer/payments/transactions#reporting-actions-for-transactions
we have a description of the field:
image
IMO This part should have a link to this section.

docs/developer/payments/transactions.mdx
IKarbowiak: > Also, what about the "webhook-part": https://docs.saleor.io/developer/extending/webhooks/synchronous-events/transaction#response-4
> Don't we need to also say there that the reference can be optional for specific events?

From what I see it's already mentioned. 
image
But looks like the list is not full, so I will extend it.

docs/developer/payments/transactions.mdx
IKarbowiak: I moved it to the part of `Reporting actions for transactions` and I think it much better right now, please verify once again

docs/developer/payments/transactions.mdx
IKarbowiak: Done

docs/developer/payments/transactions.mdx
```diff
@@ -363,6 +377,77 @@ sequenceDiagram
...
+the [`AUTHORIZATION_ADJUSTMENT`](/developer/payments/lifecycle#authorization_adjustment) event.
+
+#### Setting the `pspReference` on the `TransactionItem`
+
+The `pspReference` value in a `TransactionItem` may initially be empty if it is not provided
+during the `TransactionCreate` mutation or as a result of the `TransactionInitialize` mutation.
+
+The `transactionItem.pspReference` is updated during payment processing and reflects the `pspReference` from the most recent event received from the payment application.
+For example:
+	•	If a transaction is successful, the `pspReferenc`e will be set based on the last `SUCCESS` event from the payment provider.
```
szczecha, line 404:
```suggestion
	•	If a transaction is successful, the `pspReference` will be set based on the last `SUCCESS` event from the payment provider.
```

szczecha (approved):

PR #1392, opened 2024-11-14
Add docs about AvaTax item code

krzysztofzuraw:
Added docs on how AvaTax app is resolving item code

### DIFF
--- a/docs/developer/app-store/apps/avatax/item-code.mdx
+++ b/docs/developer/app-store/apps/avatax/item-code.mdx
@@ -1,0 +1,9 @@
+---
+sidebar_position: 5
+title: Item Code Mapping
+---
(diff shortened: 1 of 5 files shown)


docs/developer/app-store/apps/avatax/item-code.mdx
lkostrowski, line 1: does it really deserve a separate page? I think we already have some chapters where we can put this section? Adding new page adds extra maintenance cost (dead links, redirects map etc)

krzysztofzuraw: Makes sense - I moved it to `tax-code` page

lkostrowski (approved):

PR #1416, opened 2024-12-13
Transaction amount calculations

IKarbowiak:
Explain transaction amount calculations

### DIFF
--- a/docs/developer/payments/price-calculations.mdx
+++ b/docs/developer/payments/price-calculations.mdx
@@ -56,35 +56,83 @@ then the calculations are applied in the following order:
   4. recalculate amounts based on `REFUND` events
   5. recalculate amounts based on `CANCEL` events
 
-#### Authorized Value ([`transactionItem.authorizedAmount`](/api-reference/payments/objects/transaction-item#transactionitemsuthorizedamountmoney---) )
+#### Authorized Value ([`transactionItem.authorizedAmount`](/api-reference/payments/objects/transaction-item#transactionitemsuthorizedamountmoney---))
   - When an `AUTHORIZATION_SUCCESS` event occurs and there is no `AUTHORIZATION_FAILURE` or the `FAILURE` event is older: value increases by the `SUCCESS` event's amount.
   - When an `AUTHORIZATION_ADJUSTMENT` event occurs: value is overwritten by the adjustment's amount.
   - When a `CHARGE_REQUEST` or `CHARGE_SUCCESS` event occurs: value is reduced by the event's amount. 
   - When a `CANCEL_REQUEST` or `CANCEL_SUCCESS` event occurs: value is reduced by the event's amount. 
   - The amount cannot be negative, the minimum value is 0.
 
-#### Authorization Pending Value ([`transactionItem.authorizePendingAmount`](/api-reference/payments/objects/transaction-item#transactionitemauthorizependingamountmoney---) )
+#### Authorization Pending Value ([`transactionItem.authorizePendingAmount`](/api-reference/payments/objects/transaction-item#transactionitemauthorizependingamountmoney---))
   - When an `AUTHORIZATION_REQUEST` event occurs and there is no `AUTHORIZATION_FAILURE` or `AUTHORIZATION_SUCCESS` event is: value increases by the `REQUEST` event's amount.
 
-#### Charge Value ([`transactionItem.chargedAmount`](/api-reference/payments/objects/transaction-item#transactionitemchargedamountmoney---) )
+#### Charge Value ([`transactionItem.chargedAmount`](/api-reference/payments/objects/transaction-item#transactionitemchargedamountmoney---))
   - When an `CHARGE_SUCCESS` event occurs and there is no `CHARGE_FAILURE` or the `FAILURE` event is older: value increases by the `SUCCESS` event's amount.
   - When a `CHARGE_BACK` event occurs: value decreases by the event's amount.
   - When a `REFUND_REQUEST` or `REFUND_SUCCESS` event occurs: value is reduced by the event's amount. 
   - When a `REFUND_REVERSE` event occurs: value is increased by the event's amount. 
   - The value can be negative.REFUND_REVERSE
 
-#### Charge Pending Value ([`transactionItem.chargePendingAmount`](/api-reference/payments/objects/transaction-item#transactionitemchargependingamountmoney---) )
+#### Charge Pending Value ([`transactionItem.chargePendingAmount`](/api-reference/payments/objects/transaction-item#transactionitemchargependingamountmoney---))
   - When an `CHARGE_REQUEST` event occurs and there is no `CHARGE_FAILURE` or `CHARGE_SUCCESS` event is: value increases by the `REQUEST` event's amount.
 
-#### Refunded Value ([`transactionItem.refundedAmount`](/api-reference/payments/objects/transaction-item#transactionitemrefundedamountmoney---) )
+#### Refunded Value ([`transactionItem.refundedAmount`](/api-reference/payments/objects/transaction-item#transactionitemrefundedamountmoney---))
   - When an `REFUND_SUCCESS` event occurs and there is no `REFUND_FAILURE` or the `FAILURE` event is older: value increases by the `SUCCESS` event's amount.
   - When a `REFUND_REVERSE` event occurs: value decreases by the event's amount.
 
-#### Refund Pending Value ([`transactionItem.refundPendingAmount`](/api-reference/payments/objects/transaction-item#transactionitemrefundpendingamountmoney---) )
+#### Refund Pending Value ([`transactionItem.refundPendingAmount`](/api-reference/payments/objects/transaction-item#transactionitemrefundpendingamountmoney---))
   - When an `REFUND_REQUEST` event occurs and there is no `REFUND_FAILURE` or `REFUND_SUCCESS` event is: value increases by the `REQUEST` event's amount.
 
 #### Canceled Value ([`transactionItem.canceledAmount`](/api-reference/payments/objects/transaction-item#transactionitemcanceledamountmoney---) )
   - When an `CANCEL_SUCCESS` event occurs and there is no `CANCEL_FAILURE` or the `FAILURE` event is older: value increases by the `SUCCESS` event's amount.
 
-#### Cancel Pending Value ([`transactionItem.cancelPendingAmount`](/api-reference/payments/objects/transaction-item#transactionitemcancelpendingamountmoney---) )
+#### Cancel Pending Value ([`transactionItem.cancelPendingAmount`](/api-reference/payments/objects/transaction-item#transactionitemcancelpendingamountmoney---))
   - When an `CANCEL_REQUEST` event occurs and there is no `CANCEL_FAILURE` or `CANCEL_SUCCESS` event is: value increases by the `REQUEST` event's amount.
+
+
+## Total Balance Calculation
+
+The `totalBalance` represents the difference between the total amount charged (including pending charges) and the expected total cost.
+
+This balance indicates whether the customer has overpaid (positive balance) or underpaid (negative balance) for the order.
+
+Below is an explanation of how `totalBalance` is computed.
+
+### Total balance for Checkout ([`checkout.totalBalance`](/api-reference/checkout/objects/checkout#checkouttotalbalancemoney---))
+
+For a Checkout, `totalBalance` reflects the remaining balance after considering the total checkout cost, charged amounts, and pending charges.
+It is calculated as the difference between the sum of all successful and pending charges across associated transactions and the checkout’s total price:
+
+```
+totalBalance = totalCharged - checkout.totalPrice
+```
+
+where:
+```
+totalCharged = (
+    sum(transaction.chargedAmount for transaction in transactions)
+    + sum(transaction.chargePendingAmount for transaction in transactions)
+)
+```
+
+### Total balance for Order ([`order.totalBalance`](/api-reference/orders/objects/order#ordertotalbalancemoney---))
+
+For an Order, `totalBalance` represents the remaining balance after accounting for the total order cost, charged amounts, and refunds.
+It is calculated as the difference between the total charged (including pending charges) and the order cost adjusted for granted refunds:
+
+```
+totalBalance = totalCharged - (order.totalPrice - totalGrantedRefund)
+```
+
+where `totalCharged` is sum of all successful and pending charges across associated transactions:
+```
+totalCharged = (
+    sum(transaction.amountCharged for transaction in transactions)
+    + sum(transaction.amountChargePending for transaction in transactions)
+)
+```
+
+and `totalGrantedRefund` is a total of all refunds issued to the customer:
+```
+totalGrantedRefund = sum(grantedRefund.amount for grantedRefund in grantedRefunds)
+```
--- a/docs/developer/payments/price-calculations.mdx
+++ b/docs/developer/payments/price-calculations.mdx
@@ -1,0 +1,90 @@
+---
+title: Price calculations
+---
+
+This document provides an overview of how various transaction amount fields in a `TransactionItem` are calculated and influenced. Additionally, it provides details on how the totalBalance is calculated for Checkout and Order entities.
+
+## Transaction Amount Fields Overview
+
+The following fields are calculated based on transaction events and their types:
+
+- **`authorizedAmount`**: The total amount successfully authorized.
+- **`chargedAmount`**: The total amount successfully charged.
+- **`refundedAmount`**: The total amount successfully refunded.
+- **`canceledAmount`**: The total amount successfully canceled.
+- **`authorizePendingAmount`**: The total amount of pending authorizations.
+- **`chargePendingAmount`**: The total amount of pending charges.
+- **`refundPendingAmount`**: The total amount of pending refunds.
+- **`cancelPendingAmount`**: The total amount of pending cancellations.
+
+### Assumptions
+
+#### 1. Event Grouping:
+  - Events are grouped based on their PSP reference and type of action (e.g., authorization, charge).
+  - Grouping ensures that events of the same type and PSP reference are properly matched to calculate amounts accurately.
+#### 2. Pending Amounts:
+  - pending amounts (`authorizePendingAmount`, `chargePendingAmount`, `refundPendingAmount`, `cancelPendingAmount`) are increased only if a `REQUEST` event exists for the corresponding PSP reference.
+  - If a success or failure event is also associated with the same PSP reference, the system assumes the requested amount has already been processed, and the pending amount will not be increased.
+#### 3. `SUCCESS` and `FAILURE` Events:
+  - The transaction amount (e.g., `authorizedAmount`, `chargedAmount`) is increased only when a `SUCCESS` event exists.
+  - If both a `SUCCESS` and `FAILURE` event are present for the same PSP reference, the system compares their creation timestamps:
+	- If the `FAILURE` event is newer, the `SUCCESS` event is ignored.
+	- If the `SUCCESS` event is newer, it is used for the calculations.
+#### 4. Events Without PSP References:
+  - Some events (e.g., those created by the system to maintain correct amounts) may not have a PSP reference.
+  - The events of the following types are still included in the transaction amounts and influence the calculated values:
+    - `AUTHORIZATION_SUCCESS`
+    - `AUTHORIZATION_ADJUSTMENT`
+    - `CHARGE_SUCCESS`
+    - `CHARGE_BACK`
+    - `REFUND_SUCCESS`
+    - `REFUND_REVERSE`
+    - `CANCEL_SUCCESS`
+#### 5. Adjustment Events:
+  - The `AUTHORIZATION_ADJUSTMENT`, overwrite previous amounts, if an adjustment is present,
+  older authorization events are ignored, and the adjusted amount becomes the new authorized value.
+
+This logic ensures that transaction amounts reflect the current state of the transaction based on its event history while accounting for both processed and pending actions.
+
+### Calculation Details
+
+When the transaction amounts are recalculated, firstly all values are set to zero, 
+then the calculations are applied in the following order:
+  1. recalculate amounts based on events without PSP reference
+  2. recalculate amounts based on `AUTHORIZATION` events
+  3. recalculate amounts based on `CHARGE` events
+  4. recalculate amounts based on `REFUND` events
+  5. recalculate amounts based on `CANCEL` events
+
+#### Authorized Value ([`transactionItem.authorizedAmount`](/api-reference/payments/objects/transaction-item#transactionitemsuthorizedamountmoney---) )
+  - When an `AUTHORIZATION_SUCCESS` event occurs and there is no `AUTHORIZATION_FAILURE` or the `FAILURE` event is older: value increases by the `SUCCESS` event's amount.
+  - When an `AUTHORIZATION_ADJUSTMENT` event occurs: value is overwritten by the adjustment's amount.
+  - When a `CHARGE_REQUEST` or `CHARGE_SUCCESS` event occurs: value is reduced by the event's amount. 
+  - When a `CANCEL_REQUEST` or `CANCEL_SUCCESS` event occurs: value is reduced by the event's amount. 
+  - The amount cannot be negative, the minimum value is 0.
+
+#### Authorization Pending Value ([`transactionItem.authorizePendingAmount`](/api-reference/payments/objects/transaction-item#transactionitemauthorizependingamountmoney---) )
+  - When an `AUTHORIZATION_REQUEST` event occurs and there is no `AUTHORIZATION_FAILURE` or `AUTHORIZATION_SUCCESS` event is: value increases by the `REQUEST` event's amount.
+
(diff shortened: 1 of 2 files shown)


docs/developer/payments/price-calculations.mdx
```diff
@@ -0,0 +1,138 @@
...
+- **`refundPendingAmount`**: The total amount of pending refunds.
+- **`cancelPendingAmount`**: The total amount of pending cancellations.
+
+### Assumptions
+
+#### 1. Event Grouping:
+  - Events are grouped based on their PSP reference and type of action (e.g., authorization, charge).
+  - Grouping ensures that events of the same type and PSP reference are properly matched to calculate amounts accurately.
+#### 2. Pending Amounts:
+  - pending amounts (`authorizePendingAmount`, `chargePendingAmount`, `refundPendingAmount`, `cancelPendingAmount`) are increased only if a `REQUEST` event exists for the corresponding PSP reference.
```
witoszekdev, line 26: suggestion: typo

```suggestion
  - Pending amounts (`authorizePendingAmount`, `chargePendingAmount`, `refundPendingAmount`, `cancelPendingAmount`) are increased only if a `REQUEST` event exists for the corresponding PSP reference.
```

docs/developer/payments/price-calculations.mdx
```diff
@@ -0,0 +1,138 @@
...
+  - If a success or failure event is also associated with the same PSP reference, the system assumes the requested amount has already been processed, and the pending amount will not be increased.
+#### 3. `SUCCESS` and `FAILURE` Events:
+  - The transaction amount (e.g., `authorizedAmount`, `chargedAmount`) is increased only when a `SUCCESS` event exists.
+  - If both a `SUCCESS` and `FAILURE` event are present for the same PSP reference, the system compares their creation timestamps:
+	- If the `FAILURE` event is newer, the `SUCCESS` event is ignored.
+	- If the `SUCCESS` event is newer, it is used for the calculations.
```
witoszekdev, line 32: suggestion: indentation

```suggestion
		- If the `FAILURE` event is newer, the `SUCCESS` event is ignored.
		- If the `SUCCESS` event is newer, it is used for the calculations.
```

docs/developer/payments/price-calculations.mdx
```diff
@@ -0,0 +1,138 @@
...
+#### 4. Events Without PSP References:
+  - Some events (e.g., those created by the system to maintain correct amounts) may not have a PSP reference.
+  - The events of the following types are still included in the transaction amounts and influence the calculated values:
+    - `AUTHORIZATION_SUCCESS`
+    - `AUTHORIZATION_ADJUSTMENT`
+    - `CHARGE_SUCCESS`
+    - `CHARGE_BACK`
+    - `REFUND_SUCCESS`
+    - `REFUND_REVERSE`
+    - `CANCEL_SUCCESS`
```
witoszekdev, line 42: suggestion: stylistic change, so that it's clear "some events" are these listed below (I got confused by this while reading 🙈)

```suggestion
  - Not all events are required to have a PSP reference, however they still are included in the transaction amount calculation:
    - `AUTHORIZATION_SUCCESS`
    - `AUTHORIZATION_ADJUSTMENT`
    - `CHARGE_SUCCESS`
    - `CHARGE_BACK`
    - `REFUND_SUCCESS`
    - `REFUND_REVERSE`
    - `CANCEL_SUCCESS`
```

I also don't understand the part: "e.g. those created by the system to maintain correct amounts" - when such things happen? When calling `transactionUpdate`?

docs/developer/payments/price-calculations.mdx
```diff
@@ -0,0 +1,138 @@
...
+#### 5. Adjustment Events:
+  - The `AUTHORIZATION_ADJUSTMENT`, overwrite previous amounts, if an adjustment is present,
```
witoszekdev, line 44:
```suggestion
  - The `AUTHORIZATION_ADJUSTMENT`, overrides previous amount from previous `AUTHORIZATION` events, this means that if an adjustment is present,
```

docs/developer/payments/price-calculations.mdx
```diff
@@ -0,0 +1,138 @@
...
+- **`authorizedAmount`**: The total amount successfully authorized.
+- **`chargedAmount`**: The total amount successfully charged.
+- **`refundedAmount`**: The total amount successfully refunded.
+- **`canceledAmount`**: The total amount successfully canceled.
+- **`authorizePendingAmount`**: The total amount of pending authorizations.
+- **`chargePendingAmount`**: The total amount of pending charges.
+- **`refundPendingAmount`**: The total amount of pending refunds.
+- **`cancelPendingAmount`**: The total amount of pending cancellations.
+
+### Assumptions
```
witoszekdev, line 20: suggestion: I think it would be easier to read if we put Assumptions section **after** a brief explanation how does the amount calculation even works, what this is.

```suggestion
### Summary

When the transaction amounts are recalculated, firstly all values are set to zero, 
then the calculations are applied in the following order:
  1. recalculate amounts based on events without PSP reference
  2. recalculate amounts based on `AUTHORIZATION` events
  3. recalculate amounts based on `CHARGE` events
  4. recalculate amounts based on `REFUND` events
  5. recalculate amounts based on `CANCEL` events

### Assumptions
```

docs/developer/payments/price-calculations.mdx
```diff
@@ -0,0 +1,138 @@
+---
+title: Price calculations
+---
+
+This document provides an overview of how various transaction amount fields in a `TransactionItem` are calculated and influenced. Additionally, it provides details on how the totalBalance is calculated for Checkout and Order entities.
```
witoszekdev, line 5:
```suggestion
This document provides an overview of how various transaction amount fields in a `TransactionItem` are calculated and influenced. Additionally, it provides details on how `totalBalance` is calculated for Checkout and Order entities.
```

docs/developer/payments/price-calculations.mdx
```diff
@@ -0,0 +1,138 @@
...
+  3. recalculate amounts based on `CHARGE` events
+  4. recalculate amounts based on `REFUND` events
+  5. recalculate amounts based on `CANCEL` events
+
+#### Authorized Value ([`transactionItem.authorizedAmount`](/api-reference/payments/objects/transaction-item#transactionitemauthorizedamountmoney---))
+  - When an `AUTHORIZATION_SUCCESS` event occurs and there is no `AUTHORIZATION_FAILURE` or the `FAILURE` event is older: value increases by the `SUCCESS` event's amount.
+  - When an `AUTHORIZATION_ADJUSTMENT` event occurs: value is overwritten by the adjustment's amount.
+  - When a `CHARGE_REQUEST` or `CHARGE_SUCCESS` event occurs: value is reduced by the event's amount. 
+  - When a `CANCEL_REQUEST` or `CANCEL_SUCCESS` event occurs: value is reduced by the event's amount. 
+  - The amount cannot be negative, the minimum value is 0.
```
witoszekdev, line 64: suggestion: I think we should use given > when > then format for such description, and avoid using "occurs" for events, because we recalculate history on each change.

This way I thin that I've found we are missing some info in this paragraph:
- **WHEN** CHARGE_REQUEST and CHARGE_SUCCESS reduce `authorizedAmount`: I think such event must be **after** authorization and they both must have the same pspReference? I also think there cannot be `CHARGE_FAILURE` event?
- **WHEN** CANCEL_REQUEST and CANCEL_SUCCESs reduce `authorizedAmount`: same thing, it must be **after** authorization event? and there cannot be CANCEL_FAILURE event?

```suggestion
  - **GIVEN** there is an `AUTHORIZATION_SUCCESS` event
    - **WHEN there is NO** `AUTHORIZATION_FAILURE`
    - **OR WHEN** the `AUTHORIZATION_FAILURE` event is older than `AUTHORIZATION_SUCCESS`
      - **THEN** value increases by the `AUTHORIZATION_SUCCESS` event's amount.
  - **GIVEN** an `AUTHORIZATION_ADJUSTMENT` event
    - **WHEN** it's the latest event of this type
      - **THEN** value is overwritten by the `AUTHORIZATION_ADJUSTMENT` event amount.
  - **GIVEN** `CHARGE_REQUEST` or `CHARGE_SUCCESS` event:
    - **THEN** value is reduced by the event's amount. 
  - **GIVEN** `CANCEL_REQUEST` or `CANCEL_SUCCESS` event:
    - **THEN** value is reduced by the event's amount.
  - **GIVEN** all calculations were performed and we have result `authorizationAmount`
    - **WHEN** calculated `authorizationAmount` is below `0`
  	  - **THEN** `authorizationAmount` is set to `0` (it cannot be lower than 0)
```

docs/developer/payments/price-calculations.mdx
```diff
@@ -0,0 +1,138 @@
...
+#### Canceled Value ([`transactionItem.canceledAmount`](/api-reference/payments/objects/transaction-item#transactionitemcanceledamountmoney---) )
+  - When an `CANCEL_SUCCESS` event occurs and there is no `CANCEL_FAILURE` or the `FAILURE` event is older: value increases by the `SUCCESS` event's amount.
+
+#### Cancel Pending Value ([`transactionItem.cancelPendingAmount`](/api-reference/payments/objects/transaction-item#transactionitemcancelpendingamountmoney---))
+  - When an `CANCEL_REQUEST` event occurs and there is no `CANCEL_FAILURE` or `CANCEL_SUCCESS` event is: value increases by the `REQUEST` event's amount.
+
+
+## Total Balance Calculation
+
+The `totalBalance` represents the difference between the total amount charged (including pending charges) and the expected total cost.
```
witoszekdev, line 95: suggestion: swapping these two makes it more clear, that `totalBalance` is about the _amount left to be paid_

issue: "including pending charges" is not valid for Order, so maybe we shouldn't put it here? 🤔 

```suggestion
The `totalBalance` represents the difference between the expected total cost and the total amount charged (including pending charges).
```

docs/developer/payments/price-calculations.mdx
```diff
@@ -0,0 +1,138 @@
...
+```
+totalCharged = (
+    sum(transaction.chargedAmount for transaction in transactions)
+    + sum(transaction.chargePendingAmount for transaction in transactions)
+)
+```
+
+### Total balance for Order ([`order.totalBalance`](/api-reference/orders/objects/order#ordertotalbalancemoney---))
+
+For an Order, `totalBalance` represents the remaining balance after accounting for the total order cost, charged amounts, and refunds.
```
witoszekdev, line 120:
```suggestion
For an Order, `totalBalance` represents the remaining balance after accounting for the total order cost, charged amounts, and granted refunds.
```

witoszekdev, line 120: question: what is total order cost? can we use instead a field from GraphQL? is it `Order.total.gross.amount`?

docs/developer/payments/price-calculations.mdx
```diff
@@ -0,0 +1,138 @@
...
+
+#### Authorized Value ([`transactionItem.authorizedAmount`](/api-reference/payments/objects/transaction-item#transactionitemauthorizedamountmoney---))
+  - When an `AUTHORIZATION_SUCCESS` event occurs and there is no `AUTHORIZATION_FAILURE` or the `FAILURE` event is older: value increases by the `SUCCESS` event's amount.
+  - When an `AUTHORIZATION_ADJUSTMENT` event occurs: value is overwritten by the adjustment's amount.
+  - When a `CHARGE_REQUEST` or `CHARGE_SUCCESS` event occurs: value is reduced by the event's amount. 
+  - When a `CANCEL_REQUEST` or `CANCEL_SUCCESS` event occurs: value is reduced by the event's amount. 
+  - The amount cannot be negative, the minimum value is 0.
+
+#### Authorization Pending Value ([`transactionItem.authorizePendingAmount`](/api-reference/payments/objects/transaction-item#transactionitemauthorizependingamountmoney---))
+  - When an `AUTHORIZATION_REQUEST` event occurs and there is no `AUTHORIZATION_FAILURE` or `AUTHORIZATION_SUCCESS` event is: value increases by the `REQUEST` event's amount.
```
witoszekdev, line 67: question: aren't we missing here CANCEL and CHARGE events as well? I think they also lower the pending amount

docs/developer/payments/price-calculations.mdx
```diff
@@ -0,0 +1,138 @@
...
+
+#### Charge Value ([`transactionItem.chargedAmount`](/api-reference/payments/objects/transaction-item#transactionitemchargedamountmoney---))
+  - When an `CHARGE_SUCCESS` event occurs and there is no `CHARGE_FAILURE` or the `FAILURE` event is older: value increases by the `SUCCESS` event's amount.
+  - When a `CHARGE_BACK` event occurs: value decreases by the event's amount.
+  - When a `REFUND_REQUEST` or `REFUND_SUCCESS` event occurs: value is reduced by the event's amount. 
+  - When a `REFUND_REVERSE` event occurs: value is increased by the event's amount. 
+  - The value can be negative.REFUND_REVERSE
```
witoszekdev, line 74: fix: typo

```suggestion
  - This value can be negative.
```

docs/developer/payments/price-calculations.mdx
```diff
@@ -0,0 +1,138 @@
...
+
+#### Charge Pending Value ([`transactionItem.chargePendingAmount`](/api-reference/payments/objects/transaction-item#transactionitemchargependingamountmoney---))
+  - When an `CHARGE_REQUEST` event occurs and there is no `CHARGE_FAILURE` or `CHARGE_SUCCESS` event is: value increases by the `REQUEST` event's amount.
```
witoszekdev, line 77: question: isn't this value impacted by refund and chargeback as well?

docs/developer/payments/price-calculations.mdx
witoszekdev, line 74: suggestion: I would add this note, so that it's clear **this value doesn't change after CANCEL**

```suggestion
  - The value can be negative.REFUND_REVERSE
  
:::note
Please note that `chargedAmount` isn't impacted by `CANCEL` events - these are only meant for matching with `AUTHORIZATION` event
:::
```

docs/developer/payments/price-calculations.mdx
```diff
@@ -0,0 +1,138 @@
...
+
+#### Refunded Value ([`transactionItem.refundedAmount`](/api-reference/payments/objects/transaction-item#transactionitemrefundedamountmoney---))
```
witoszekdev, line 79: question: doesn't CHARGEBACK event also increase `refundedAmount`? if not then we can add a note:

```
:::note
Note that `CHARGEBACK` events which work similarly to a refund (decreases chargedAmount value), are not included in **refundedAmount**, since they were not triggered by merchant, but billed by issuing bank.
:::
```

docs/developer/payments/price-calculations.mdx
```diff
@@ -0,0 +1,138 @@
...
+  - When an `REFUND_REQUEST` event occurs and there is no `REFUND_FAILURE` or `REFUND_SUCCESS` event is: value increases by the `REQUEST` event's amount.
+
+#### Canceled Value ([`transactionItem.canceledAmount`](/api-reference/payments/objects/transaction-item#transactionitemcanceledamountmoney---) )
+  - When an `CANCEL_SUCCESS` event occurs and there is no `CANCEL_FAILURE` or the `FAILURE` event is older: value increases by the `SUCCESS` event's amount.
+
+#### Cancel Pending Value ([`transactionItem.cancelPendingAmount`](/api-reference/payments/objects/transaction-item#transactionitemcancelpendingamountmoney---))
+  - When an `CANCEL_REQUEST` event occurs and there is no `CANCEL_FAILURE` or `CANCEL_SUCCESS` event is: value increases by the `REQUEST` event's amount.
+
+
+## Total Balance Calculation
```
witoszekdev, line 93: question: maybe we should put it in **Payment lifecycle** docs page, since it also has explanation for `authorizeStatus` and `chargeStatus`, so this would nicely fit into these docs? (they also have formulas that explain it similarly to this description).

I have PR open to add more 😄 
https://github.com/saleor/saleor-docs/pull/1379/files

docs/developer/payments/price-calculations.mdx
IKarbowiak: The pending amount are also include in order calculations for `total_balance`.

docs/developer/payments/price-calculations.mdx
IKarbowiak: No, `CHARGEBACK` does not influence `charge_pending_amount`

docs/developer/payments/price-calculations.mdx
IKarbowiak: Note added

docs/developer/payments/price-calculations.mdx
IKarbowiak: Done

docs/developer/payments/price-calculations.mdx
```diff
@@ -0,0 +1,179 @@
...
+
+### Assumptions
+
+#### 1. Event Grouping:
+  - Events are grouped based on their PSP reference and type of action (e.g., authorization, charge).
+  - Grouping ensures that events of the same type and PSP reference are properly matched to calculate amounts accurately.
+#### 2. Pending Amounts:
+  - Pending amounts (`authorizePendingAmount`, `chargePendingAmount`, `refundPendingAmount`, `cancelPendingAmount`) are increased only if a `REQUEST` event exists for the corresponding PSP reference.
+  - If a success or failure event is also associated with the same PSP reference, the system assumes the requested amount has already been processed, and the pending amount will not be increased.
+#### 3. `SUCCESS` and `FAILURE` Events:
```
korycins, line 38: Wouldn't be also good to mention that in case of the SUCCESS, the amount will be subtracted from the "previous" value? 
I mean, we have a transaction with event `AUTHORIZE_SUCCESS` and amount 100. TR has authorizedAmount:100, chargedAmount:0. Then we receive the event CHARGE_SUCESS and amount: 100. The 100 will have an impact on authorizedAmount. TR will have: authorizedAmount:0, chargedAmount:100 (as we subtracted the amount from CHARGE_SUCCESS from authorizedAmount and added it to chargedAmount)

korycins: Okay, I see that you deeply explained it below. So the question is: Do you think, that this is wort also to mention it here?

docs/developer/payments/price-calculations.mdx
```diff
@@ -0,0 +1,179 @@
...
+    - `REFUND_SUCCESS`
+    - `REFUND_REVERSE`
+    - `CANCEL_SUCCESS`
+#### 5. Adjustment Events:
+  - The `AUTHORIZATION_ADJUSTMENT`, overrides previous amount from previous `AUTHORIZATION` events, this means that if an adjustment is present,
+  older authorization events are ignored, and the adjusted amount becomes the new authorized value.
+
+This logic ensures that transaction amounts reflect the current state of the transaction based on its event history while accounting for both processed and pending actions.
+
+### Calculation details
```
korycins, line 58: Maybe worth explicitly say, that we are talking about the events returned by APP - I mean, events that have `psp_reference` provided by App. For example, you could still see `X_request` event created by Saleor when calling the app, but it is not included in calculations until receiving at least psp_refernce from payment app 🤔

docs/developer/payments/price-calculations.mdx
```diff
@@ -0,0 +1,179 @@
...
+      - **THEN** value increases by the `AUTHORIZATION_REQUEST` event's amount
+
+#### Charge Value ([`transactionItem.chargedAmount`](/api-reference/payments/objects/transaction-item#transactionitemchargedamountmoney---))
+  - **GIVEN** there is a `CHARGE_SUCCESS` event
+    - **WHEN** there is no `CHARGE_FAILURE`
+    - **OR WHEN** the `CHARGE_FAILURE` event is older than `CHARGE_SUCCESS`
+      - **THEN** value increases by the `CHARGE_SUCCESS` event's amount.
+  - **GIVEN** there is a `CHARGE_BACK` event
+    - **THEN**value is reduced by the event's amount
+  - **GIVEN** there is a `REFUND_REQUEST` or `CHARGE_SUCCESS` event:
```
korycins, line 88: Shoudln't be `REFUND_SUCCESS`? 🤔

docs/developer/payments/lifecycle.mdx
```diff
@@ -359,6 +359,8 @@ Each event can be excluded from calculation if it has matching `*_FAILURE` event
 
 To learn more about each event type, see [Transaction event types](#transaction-event-types) section.
 
+To learn more about amount calculations see [Price calculations](/developer/payments/price-calculations) page.
```
witoszekdev, line 362: We need `.mdx` ending so that file renaming works correctly 🤓

```suggestion
To learn more about amount calculations see [Price calculations](/developer/payments/price-calculations.mdx) page.
```

witoszekdev (dismissed):

docs/developer/payments/price-calculations.mdx
korycins: I might not be clear here. 
I was meaning the situation when saleor creates X_REQUEST event (like calling charge-request, or refund-request). The event's amount will not be included in calculations until receiving the SUCCESS response from the app. As this action is called asynchronously (I mean a situation when calling this action from dashboard), we can have a TR like:
TR: chargedAmount 100, refundAmount:0
Event1: amount: 100, type CHARGE_SUCCESS, pspRef:123
Event2: amount: 100, type: REFUND_REQUEST, pspRef:null 

As event2 was created by Saleor, and we're proceeding the refund request right now, the transaction.chargedAmount , transaction.refundedAmount are not impacted.

docs/developer/payments/price-calculations.mdx
IKarbowiak: I added a note about that in `Calculation process` part

docs/developer/payments/price-calculations.mdx
IKarbowiak: I added a note about that

witoszekdev (approved):

docs/developer/payments/price-calculations.mdx
```diff
@@ -0,0 +1,185 @@
...
+#### 2. Pending Amounts:
+  - Pending amounts (`authorizePendingAmount`, `chargePendingAmount`, `refundPendingAmount`, `cancelPendingAmount`) are increased only if a `REQUEST` event exists for the corresponding PSP reference.
+  - If a success or failure event is also associated with the same PSP reference, the system assumes the requested amount has already been processed, and the pending amount will not be increased.
+#### 3. `SUCCESS` and `FAILURE` Events:
+  - The transaction amount (e.g., `authorizedAmount`, `chargedAmount`) is increased only when a `SUCCESS` event exists.
+  - If both a `SUCCESS` and `FAILURE` event are present for the same PSP reference, the system compares their creation timestamps:
+    - If the `FAILURE` event is newer, the `SUCCESS` event is ignored.
+    - If the `SUCCESS` event is newer, it is used for the calculations.
+#### 4. Events Without PSP References:
+  - Not all events require a PSP reference, but they are still included in the transaction amount calculation. The following events do not need a PSP reference:
```
korycins, line 46: IMHO this could be misleading, We have two ways of handling the transactions:
a) TransactionInitialize & TransactionProcess & transactionEventReport - flow which can be embedded in checkout process, where Saleor is a "proxy" between front and payment-app
b) TransactionCreate & TransactionUpdate - protected mutation that allows you to create and modify the transaction. 

The `a)` always requires the psp-reference as the whole calculation process requires it. For example, we don't allow the app to not return the psp-reference with success.  The whole flow relies on the psp-refences.

For `b)` when calling `transactionCreate` with chargedAmount:100,  we create the new events: to be able to re-calculate the amounts, but it is our inside logic. As a result of `transactionCreate` mutation, you expect to get a proper values in `transaction.xAmount` fields. The events without psp-reference is the "side-effect" of this action, as we want to be able to re-calculate the values. My point here is that we don't offer any API that would allow you to provide the event explicitly without psp-refence, which would be included in calculations.

So IMHO listing the `X_SUCCESS` events could be misleading, as someone could assume that `a)` flow would not require the psp-refence (which is not true). 

Maybe we could consider changing the section name, to say that in case of using `transactionUpdate`/`transactionCreate`, the events will not have a psp-refernce, and how we handle them 🤔

docs/developer/payments/price-calculations.mdx
```diff
@@ -0,0 +1,185 @@
...
+
+This logic ensures that transaction amounts reflect the current state of the transaction based on its event history while accounting for both processed and pending actions.
+
+### Calculation details
+:::note
+Events like `X_REQUEST`, generated by Saleor when calling the payment app, are excluded from calculations until a corresponding event with a valid `psp_reference` is received from the payment app.
+For example, an `AUTHORIZATION_REQUEST` event will not affect the calculations until an `AUTHORIZATION_SUCCESS` or a similar event containing a `psp_reference` is returned by the payment app.
+:::
+
+#### Authorized Value ([`transactionItem.authorizedAmount`](/api-reference/payments/objects/transaction-item#transactionitemauthorizedamountmoney---))
```
korycins, line 66: Maybe worth mentioning that the below calculation flows are for the events with the same psp-refence?
For example:
```
- **GIVEN** there is a `CHARGE_SUCCESS` event
     - **WHEN** there is no `CHARGE_FAILURE`
```
We can have `CHARGE_SUCCESS` with PSP:123, and `CHARGE_FAILURE` with PSP:ABC. In that case FAILURE will not have any impact on SUCCESS, as PSP is different

korycins (approved):

witoszekdev (approved):

PR #1435, opened 2025-01-10
Add Segment app docs

krzysztofzuraw:


docs/developer/app-store/apps/segment.mdx
```diff
@@ -0,0 +1,98 @@
...
+- `ORDER_CANCELLED`
+- `ORDER_FULLY_PAID`
+- `ORDER_REFUNDED`
+- `ORDER_UPDATED`
+
+For each webhook, the application will send the following information to Segment
+
+- Order `id` in base64 format
+- User information:
+    - User `id` in base64 format
```
witoszekdev, line 47: question: Is this exactly `Order.id` from Saleor or do we do some additional `base64` encoding?

I'm wondering if we should even mention it's base64, since it can change at any time, because Saleor didn't guarantee that it will use this specific format and we **definitely** don't want to recommend users to decode that base64 because they will start depending on it.

krzysztofzuraw: exactly `Order.id` 😄 

Good question - maybe just `Order.id` is enough

lkostrowski (approved):

docs/developer/app-store/apps/segment.mdx
krzysztofzuraw: Done in [`572718b` (#1435)](https://github.com/saleor/saleor-docs/pull/1435/commits/572718b6585d8628c1cf645f93e210f8a8e6b945)

lkostrowski (approved):

PR #1447, opened 2025-01-21
Fix segment docs: diagram and event used for testing

krzysztofzuraw:


docs/developer/app-store/apps/segment.mdx
```diff
@@ -67,8 +67,8 @@ Example of successful handling of `ORDER_CREATED` webhook:
 ```mermaid
 sequenceDiagram
     OrderCreatedWebhook->>+App: Send webhook payload
-    App-->>-AppMetadata: Fetch app configuration from metadata
-    AppMetadata-->>+App: Return configuration
+    App-->>-AppConfiguration: Fetch app configuration
+    AppConfiguration-->>+App: Return configuration
```
lkostrowski, line 71: I think this is a wrong abstract. Configuration is something internal to app, contrary to previous approach where metadata *was* something external.

I suggest to change AppConfiguration to `Database` and thats it - configuration is indeed fetched from DB
The details about DynamoDB can be left to "development" part

lkostrowski (approved):

PR #1454, opened 2025-01-27
Add missing compatibility info to CMS app Strapi config

Cloud11PL:


poulch (approved):
LGTM, but I'd think also about put this to `Main assumptions & limitations` section?

PR #1460, opened 2025-02-03
Enhance `applyOncePerOrder` voucher behaviour description

zedzior:
This PR clarifies, that `applyOncePerOrder` voucher reduces the value of a single unit, and not the entire line

### DIFF
--- a/docs/developer/discounts/vouchers.mdx
+++ b/docs/developer/discounts/vouchers.mdx
@@ -24,6 +24,7 @@ We can also distinguish vouchers by their value type:
 - `PERCENTAGE`: reduces the price by a specified percentage value.
 
 By setting `Voucher.applyOncePerOrder`, there is an option to apply the discount only to the cheapest eligible product.
+The discount reduces the value of a single unit, not the entire line.
 If the voucher specifies certain products, the discount will be applied only to the cheapest item included in the discount.
 If the voucher applies to all products, the discount will be applied only to the cheapest item overall.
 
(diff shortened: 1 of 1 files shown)


docs/developer/discounts/vouchers.mdx
```diff
@@ -24,6 +24,7 @@ We can also distinguish vouchers by their value type:
 - `PERCENTAGE`: reduces the price by a specified percentage value.
 
 By setting `Voucher.applyOncePerOrder`, there is an option to apply the discount only to the cheapest eligible product.
+The discount reduces the value of a single unit, not the entire line.
```
korycins, line 27:
```suggestion
The discount reduces the price of a single quantity of the cheapest product.
```
Maybe something like this ? ☝️ I don't think that we need to explicitly say, that this voucher is not applied on entire line. I mean, this is already clear from first part of the sentence.
I've also changed the unit into quantity, as we are using quantity in the OrderLine, CheckoutLine, so I assumed that this would better fit here. 
But this is only the suggestions 😄

korycins (approved):

docs/developer/discounts/vouchers.mdx
zedzior: Thank you for the suggestion. The purpose of this sentence is to emphasize that the this voucher is not applied to the entire line (like the all other line-level discounts), so I would keep it. I used "single unit" as I am not sure if "single quantity" phrase exists in english. Please note, that the info, that it is applied to the cheapest product is contained in the sentence right before. This my rationale, let's wait for other reviewers and let them decide what fits here better.

IKarbowiak: Maybe:
```The discount applies to a single unit of the product, not the entire order line.`

IKarbowiak (approved):

fowczarek (approved):

PR #1464, opened 2025-02-07
Update docs about Shipping events

krzysztofzuraw:


docs/developer/extending/webhooks/synchronous-events/shipping.mdx
```diff
@@ -47,6 +47,16 @@ shipping methods are queried for a checkout object.
 }
 ```
 
+:::warning
+Due to circular reference, subscription queries should not contain fields:
+
+- `checkout.shippingMethods`
+- `checkout.deliveryMethod`
```
lkostrowski, line 54: below in docs you mentioned `order` fields too, I guess it should be mentioned?

krzysztofzuraw: This is `SHIPPING_LIST_METHODS_FOR_CHECKOUT` event so it won't be order here - Saleor don't yet have `SHIPPING_LIST_METHODS_FOR_ORDER` 😞

lkostrowski (approved):

PR #1478, opened 2025-02-25
Add `Price Freeze Period` section.

zedzior:
This PR covers changes introduced in https://github.com/saleor/saleor/pull/17160

### DIFF
--- a/docs/upgrade-guides/core/3-20-to-3-21.mdx
+++ b/docs/upgrade-guides/core/3-20-to-3-21.mdx
@@ -291,3 +291,17 @@ The `external-shipping-method-id` is no longer stored within `checkout.privateMe
 Previously, the `checkoutDeliveryMethodUpdate` and `checkoutShippingMethodUpdate` mutations would trigger the `CHECKOUT_UPDATED` webhook and invalidate taxes regardless of whether the same delivery method was already assigned to the checkout.
 Starting from version 3.21, if the same method that is already assigned to the checkout is provided, the `CHECKOUT_UPDATED` webhook will not be triggered, and taxes will not be invalidated.
 From Saleor's perspective, nothing has changed, as the checkout remains in the same state as it was before the mutation was called.
+
+
+## Introduced price freeze period for editable orders
+In previous versions, all editable orders (those with `DRAFT` and `UNCONFIRMED` statuses) refreshed their base prices
+by fetching them from the current product catalog during each recalculation. 
+Additionally, all discounts were applied based on the latest reward values.
+
+To ensure price consistency over time, regardless of updates to the product catalog, Saleor has introduced a price freeze period. 
+During this period, order lines use denormalized base prices instead of retrieving them from the catalog. 
+More details can be found [`here`](developer/price-freeze-period.mdx)  
+
+- Draft orders: the base prices are frozen for 24 hours by default. This duration can be modified using the `Channel.draftOrderLinePriceFreezePeriod` setting.
+- Unconfirmed orders: the base prices remain frozen indefinitely, and this setting is not configurable.
+- Other orders: remain unchanged and cannot be recalculated.
--- a/sidebars/core-concepts.js
+++ b/sidebars/core-concepts.js
@@ -39,6 +39,7 @@ export const coreConcepts = [
   "developer/order/order-expiration",
   "developer/order/order-to-checkout",
   "developer/order/address",
+  "developer/price-freeze-period",
 
   title("Transactions"),
   "developer/payments/overview",
--- a/docs/developer/price-freeze-period.mdx
+++ b/docs/developer/price-freeze-period.mdx
@@ -1,0 +1,41 @@
+---
+title: Price Freeze Period
+sidebar_label: Price Freeze Period
+---
+
+:::info
+This feature was introduced in **Saleor 3.21**. In previous versions, the base prices were refreshed with every single recalculation.
+:::
+
+## Introduction
+The base price of a product is determined at the moment it is added to an order. 
+However, product prices may change throughout the order lifecycle. 
+To maintain price consistency over time, regardless of updates to the product catalog, Saleor provides a price freeze period.
+
+
+### Draft Orders
+For draft orders, Saleor retains the initial base price of an order line for 24 hours by default.
+After this period, the base price is considered outdated and will be refreshed based on the current product catalog.
+
+The default freeze period is controlled by the [`channel`](developer/channels/configuration.mdx) setting `draftOrderLinePriceFreezePeriod`. 
+To disable automatic price refreshing, set this value to `null` or `0`.
+
+
+### Unconfirmed Orders
+Unconfirmed orders — those originating from checkout — do not refresh their base prices at any point.
+
+
+### Non-Editable Orders
+Orders with statuses other than `DRAFT` and `UNCONFIRMED` are not editable so can't be recalculated. 
+
+
+### How It Works
+When a product is added to an order, Saleor sets an expiration time for the order line. 
+Once this period has passed, the line's base price will be updated based on the current product price.
+This update also reflects any changes in applied discounts. 
+For example, if a voucher reduced the line price and the voucher value has changed since the initial calculation, 
+the line price will be adjusted based on the new voucher value.
+
+
+### Exceptions
+- The price will not expire if the line price was [overridden](api-reference/orders/inputs/order-line-create-input.mdx#orderlinecreateinputpricepositivedecimal--).


docs/developer/price-expiration.mdx
```diff
@@ -0,0 +1,30 @@
...
+When adding a line to an order, the product variant price is taken from the relevant channel listing. 
+However, throughout the order lifecycle, price recalculations may be required, and channel listing prices can be updated in the meantime.
+To ensure price consistency over time, regardless of channel listing updates, Saleor provides a settings that allows to control the price freeze period.
+
+### Settings
+You can configure the duration after which a price is considered outdated and requires refreshing from the channel listings. 
+This is managed through the [`channel`](developer/channels/configuration.mdx) settings:
+- `draftOrderLinePriceFreezePeriod`:  Defines the time (in hours) after which the order line price will be refreshed. The default value is 24 hours. 
+To prevent price refreshing entirely, set this value to null or 0.
+- `checkoutLinePriceFreezePeriod`: To be added. Currently, each recalculation retrieves the latest price from the channel listing.
```
korycins, line 17: This could be misleading, as you provide the `checkoutLinePriceFreezePeriod`, which does not exist. Personally, I would add a separate task about updating this section when we introduce the same flow for the checkout. 
But I would keep the sentence about checkout part: 

> Currently, each recalculation retrieves the latest price from the channel listing.

Also IMHO, would be goot to mention that this flow was added in 3.21

korycins (commented):
Would be also good, to have a migration guide 3.20 -> 3.21 where you will explain, what has been changed: 
- Previously the refresh prices was happening for Draft and UNCONFIRMED, now it happens only for DRAFT
- The DRAFT prices are freeze for 24h by default, which means that we change the behaviour, so would be good to mention it.

docs/developer/price-expiration.mdx
```diff
@@ -0,0 +1,32 @@
...
+---
+
+:::info
+This feature was introduced in **Saleor 3.21**. In previous versions, the base prices were refreshed with every single recalculation.
+:::
+
+## Introduction
+When adding a line to an order, the product variant price is taken from the relevant channel listing. 
+However, throughout the order lifecycle, price recalculations may be required, and channel listing prices can be updated in the meantime.
+To ensure price consistency over time, regardless of channel listing updates, Saleor provides a settings that allows to control the price freeze period.
```
IKarbowiak, line 13: I have a trouble to understand what do you want to tell us.
`price recalculations may be required` - because of what?
`and channel listing prices can be updated in the meantime.` - how it's related with other part of the sentence?

> To ensure price consistency over time, regardless of channel listing updates, Saleor provides a settings that allows to control the price freeze period.

I don't think that mentioning here channel listings update is important. 
I think that introduction should say what happened in general, not going into the details

docs/developer/price-expiration.mdx
```diff
@@ -0,0 +1,32 @@
...
+
+### Settings
+You can configure the duration after which a price is considered outdated and requires refreshing from the channel listings.
```
IKarbowiak, line 16: Maybe:
```suggestion
You can configure the time period after which a price is considered outdated and needs to be refreshed from the channel listings.
```

docs/developer/price-expiration.mdx
```diff
@@ -0,0 +1,32 @@
...
+## Introduction
+When adding a line to an order, the product variant price is taken from the relevant channel listing. 
+However, throughout the order lifecycle, price recalculations may be required, and channel listing prices can be updated in the meantime.
+To ensure price consistency over time, regardless of channel listing updates, Saleor provides a settings that allows to control the price freeze period.
+
+### Settings
+You can configure the duration after which a price is considered outdated and requires refreshing from the channel listings. 
+This is managed through the [`channel`](developer/channels/configuration.mdx) setting `draftOrderLinePriceFreezePeriod`, 
+which defines the time in hours after which the order line price will be refreshed. The default value is 24 hours. 
+To prevent price refreshing entirely, set this value to `null` or `0`.
```
IKarbowiak, line 19:
```suggestion
To disable automatic price refreshing, set this value to `null` or `0`.
```

docs/developer/price-expiration.mdx
```diff
@@ -0,0 +1,32 @@
...
+
+## Introduction
+When adding a line to an order, the product variant price is taken from the relevant channel listing. 
+However, throughout the order lifecycle, price recalculations may be required, and channel listing prices can be updated in the meantime.
+To ensure price consistency over time, regardless of channel listing updates, Saleor provides a settings that allows to control the price freeze period.
+
+### Settings
+You can configure the duration after which a price is considered outdated and requires refreshing from the channel listings. 
+This is managed through the [`channel`](developer/channels/configuration.mdx) setting `draftOrderLinePriceFreezePeriod`, 
+which defines the time in hours after which the order line price will be refreshed. The default value is 24 hours.
```
IKarbowiak, line 18:
```suggestion
This is controlled by the [`channel`](developer/channels/configuration.mdx) setting `draftOrderLinePriceFreezePeriod`. It defines how long the price remains fixed before being refreshed, measured in hours. The default setting is 24 hours.
```

korycins (dismissed):

docs/developer/price-expiration.mdx
```diff
@@ -0,0 +1,32 @@
...
+
+### Settings
+You can configure the duration after which a price is considered outdated and requires refreshing from the channel listings. 
+This is managed through the [`channel`](developer/channels/configuration.mdx) setting `draftOrderLinePriceFreezePeriod`, 
+which defines the time in hours after which the order line price will be refreshed. The default value is 24 hours. 
+To prevent price refreshing entirely, set this value to `null` or `0`.
+
+### How It Works
+The price expiration mechanism applies only to orders with `DRAFT` status. Orders in any other status will retain their original prices permanently.
+A new price expiration time is set on an order line when:
```
IKarbowiak, line 23: Do not use `new` - how long it's gonna be `new`? Will you remember to change it after it won't being `new`? 😄 
```suggestion
The price expiration time is set on an order line when:
```

docs/developer/price-expiration.mdx
```diff
@@ -0,0 +1,32 @@
...
+- A variant is added to the order.
+- The freeze period has passed, and the line undergoes recalculation.
```
IKarbowiak, line 25: `and the line undergoes recalculation.` - what do you mean exactly?

docs/developer/price-expiration.mdx
```diff
@@ -0,0 +1,32 @@
...
+
+Once the price freeze period expires, Saleor retrieves the latest variant price from the channel listing and updates it accordingly. 
+This update also includes any changes in applied discounts. 
+For example, if a voucher reduced the line price and the voucher reward value has changed since the initial calculation, the line price will be adjusted based on the new reward value.
```
IKarbowiak, line 29:
```suggestion
Once the price freeze period expires, Saleor retrieves the latest variant price from the channel listing and recalculates the order line price. This update also reflects any changes in applied discounts.
For example, if a voucher initially reduced the line price but its value has changed since the initial calculation, the line price will be adjusted accordingly.
```

sidebars/core-concepts.js
```diff
@@ -71,6 +71,7 @@ export const coreConcepts = [
   "developer/taxes",
   "developer/thumbnails",
   "developer/price-calculation",
+  "developer/price-expiration",
```
IKarbowiak, line 74: It regards just orders, right? So maybe it should go to `Order` section instead

docs/developer/price-expiration.mdx
zedzior: corrected

docs/developer/price-expiration.mdx
zedzior: corrected

docs/developer/price-expiration.mdx
zedzior: corrected

docs/developer/price-expiration.mdx
zedzior: corrected

docs/developer/price-expiration.mdx
zedzior: > For example, if a voucher initially reduced the line price but its value has changed since the initial calculation, the line price will be adjusted accordingly.

With suggested changes, I don't get what `its value` refers to and what `the line price will be adjusted accordingly` means, so corrected partially.

docs/developer/price-expiration.mdx
zedzior: > I don't think that mentioning here channel listings update is important.

This is the essence of this change - we don't fetch prices from channel listing, but we use denormalized ones.

> price recalculations may be required - because of what?

 As you said, this is introduction so it shouldn't got much into details what can trigger the price recalculation

sidebars/core-concepts.js
IKarbowiak: It will work in the same way? If not, I think it's better to create new page for checkout and put it in the checkout section

docs/developer/price-expiration.mdx
```diff
@@ -0,0 +1,32 @@
...
+To ensure price consistency over time, regardless of channel listing updates, Saleor provides a settings that allows to control the price freeze period.
+
+### Settings
+You can configure the time period after which a price is considered outdated and needs to be refreshed from the channel listings. 
+This is controlled by the [`channel`](developer/channels/configuration.mdx) setting `draftOrderLinePriceFreezePeriod`. 
+It defines how long the price remains fixed before being refreshed, measured in hours. The default setting is 24 hours.
+To disable automatic price refreshing, set this value to `null` or `0`.
+
+### How It Works
+The price expiration mechanism applies only to orders with `DRAFT` status. Orders in any other status will retain their original prices permanently.
```
korycins, line 22: I would suggest three separate sub-sections. First for `DRAFT`, second  for `UNCONFIRMED` and the last that would explicitly say that we don't modify the prices.

Right now, you have a section `###Settings`. You explain there that you can control the expiration time for orders.
Later in section `How it works` you're pointing that the price expiration works only for DRAFT. So for someone who is not familiar this can be hard to understand (for example, does the `settings` part is related to ALL orders? or related to only specific orders? We don't say that explicit).

So IMHO, to keep it clear, I would expect the section for DRAFT, with the info that you already provided.
Second for UNCONFIRMED, where you say: we don't update the prices for existing order's content, but we will use the actual price for channel when you add the new line
Thrid one would just say: order is confirmed so we don't make any calculations.

docs/upgrade-guides/core/3-20-to-3-21.mdx
```diff
@@ -291,3 +291,17 @@ The `external-shipping-method-id` is no longer stored within `checkout.privateMe
 Previously, the `checkoutDeliveryMethodUpdate` and `checkoutShippingMethodUpdate` mutations would trigger the `CHECKOUT_UPDATED` webhook and invalidate taxes regardless of whether the same delivery method was already assigned to the checkout.
 Starting from version 3.21, if the same method that is already assigned to the checkout is provided, the `CHECKOUT_UPDATED` webhook will not be triggered, and taxes will not be invalidated.
 From Saleor's perspective, nothing has changed, as the checkout remains in the same state as it was before the mutation was called.
+
+
+## Introduced price freeze period for editable orders
+In previous versions, all editable orders (those with `DRAFT` and `UNCONFIRMED` status) refreshed their prices by retrieving them from current channel listings during each recalculation.
```
korycins, line 297: I would consider re-naming the `channel-listing` words to something more meaningful and not related to the implementation or API. Something like: `price for channel` or `actual price for channel`

IKarbowiak (approved):

korycins (approved):

PR #1485, opened 2025-02-28
Update docker-compose.mdx

arthru:
The docker compose file has nothing to build so I suggest to replace `docker compose build` by `docker compose pull` in the doc


aniav (approved):
I checked the changes and indeed build doesn't do anything.

PR #1486, opened 2025-03-01
Fix line breaks in order state diagram

rkashapov:


aniav:
Thank you for noticing this and for the changes. The file with order state has been moved in the repo so I couldn't fix the conflicts here. I created #1495 instead to address this and so that you don't need to update your PR 🙌

PR #1513, opened 2025-03-26
Mention the removed plugins

patrys:
Also fix preformatted text in section headers.

### DIFF
--- a/src/css/primitives/typography.css
+++ b/src/css/primitives/typography.css
@@ -8,7 +8,7 @@
   --font-size-xxl: clamp(2.44rem, 2.38vw + 1.85rem, 3.75rem);
   --font-size-xxxl: clamp(3.05rem, 3.54vw + 2.17rem, 5rem);
 
-  --ifm-code-font-size: 0.875rem;
+  --ifm-code-font-size: 0.875em;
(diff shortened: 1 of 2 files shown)


maarcingebala (dismissed):

aniav (dismissed):

lkostrowski (approved):

src/css/primitives/typography.css
```diff
@@ -8,7 +8,7 @@
   --font-size-xxl: clamp(2.44rem, 2.38vw + 1.85rem, 3.75rem);
   --font-size-xxxl: clamp(3.05rem, 3.54vw + 2.17rem, 5rem);
 
-  --ifm-code-font-size: 0.875rem;
+  --ifm-code-font-size: 0.875em;
```
krzysztofzuraw, line 11: **Question:** why we change to `em`? Won't it mean that it will change according to parent font-size which can lead to some inconsistencies?

patrys: This is by design. Take a look at the current headers in this guide (and API references and any other place that uses code in headers); they look illegible.

aniav (approved):

krzysztofzuraw (approved):

PR #1515, opened 2025-03-28
Add observability docs

maarcingebala:
Related: https://github.com/saleor/saleor/pull/17456


docs/setup/telemetry.mdx
```diff
@@ -0,0 +1,113 @@
...
+
+Context propagation between Saleor and external services is handled through OpenTelemetry's [W3C Trace Context](https://www.w3.org/TR/trace-context/) standard.
+
+## Setting up your own observability
+
+If you're running your own Saleor instance, you can set up telemetry collection and viewing using standard OpenTelemetry tooling. Below is an example of a observability setup where Saleor is configured to send telemetry data to an [OpenTelemetry Collector](https://opentelemetry.io/docs/collector/), which then exports the data to a [Jaeger](https://www.jaegertracing.io/) instance for visualization.
+
+### 1. Configure an OpenTelemetry Collector
+
+Set up an OpenTelemetry Collector to receive, process, and export telemetry data.
```
korycins, line 65: Would be good to mention, that this is a part of docker-compose file, right?

docs/setup/telemetry.mdx
```diff
@@ -0,0 +1,113 @@
...
+
+Note: Only traces are exported in this example as Jaeger doesn't support metrics. You can add additional exporters for metrics if needed.
+
+### 2. Set Up Environment Variables
+
+Configure the necessary environment variables for your Saleor instance to connect to your telemetry system:
+
+```bash
+# OpenTelemetry configuration
+export OTEL_SERVICE_NAME=saleor
```
korycins, line 101: Can you provide a description of the envs? It would give you mroe context than example value like `saleor` or `localhost`.

korycins: Also, don't you need to mention them here: https://docs.saleor.io/setup/configuration ?

docs/setup/telemetry.mdx
maarcingebala: When running a collector, this is a standard config file that you need. I can make it more explicit.

docs/setup/telemetry.mdx
maarcingebala: That's a good point!

docs/setup/telemetry.mdx
```diff
@@ -0,0 +1,113 @@
+---
+title: Telemetry in Saleor
+description: Learn how Saleor uses OpenTelemetry for observability and monitoring
+---
+
+# Telemetry in Saleor
+
+## Introduction
+
+Saleor implements telemetry using the OpenTelemetry framework to provide comprehensive observability of your application. This document describes how telemetry is implemented, how to configure it, and best practices for using it in your own deployment.
```
Magdalena-Kobylanska, line 10: I'm not a fan of this introduction (smells of machine :D), let me propose a different one:

Saleor uses OpenTelemetry standard to give you visibility into what's happening in Saleor Core, and between Core and apps. You can export this data into your prefered observability tool, and create traces, metrics, and monitors to suit your needs.
In this guide, you'll find information on how to implement telemetry, how to configure it, and how you can start using it in your environment.

Magdalena-Kobylanska (changes requested):

korycins (commented):
AFAIK, upgrading to a version that includes your changes will cause services like Jaeger to stop working. Any such change should be explicitly mentioned in the migration guide.

maarcingebala:
@korycins Jaeger integration will be updated separately but before the 3.21 release - we're tracking this in Linear.

docs/setup/telemetry.mdx
maarcingebala: I added a link to a page that explains the configuration. Also, I decided not to include it in the `https://docs.saleor.io/setup/configuration` page as these are not Saleor-specific settings. They can be set with a command-line argument as well, and there are many more config options available. I think it's beyond the scope of our docs, and OpenTelemetry docs should be used for that instead.

docs/setup/telemetry.mdx
maarcingebala: Added info that this is collector config and a link to docs on configuring collectors.

korycins (dismissed):

przlada (dismissed):

Magdalena-Kobylanska (dismissed):

PR #1524, opened 2025-04-07
Rewrite the benefits of composable architectures

patrys:


peelar (dismissed):
I know these articles were written to be standalone and I like them that way but maybe we could find ways to link to some other resources. 

Here is a crude example:

> We believe for larger companies, a more modular architecture is a better approach. If monolithic software is a single program, composable architectures are like an operating system with multiple programs running. Each specializes in its area of responsibility and delegates other tasks to the remaining systems.
>
> **We designed [Saleor's app architecture](link) to express just that.**

peelar (approved):

mirekm (approved):

PR #1532, opened 2025-04-11
Improve the extensibility section of "Why Saleor"

patrys:


docs/overview/why-saleor/extensibility.mdx
```diff
@@ -1,48 +1,81 @@
...
-  to receive data from Saleor.
+But before diving into the details of those aspects, we must clarify one thing.
 
-- **API extensions.**
-  Enable behavior customization, for example, calculating shipping costs based on the package dimensions.
-  [Learn more](developer/extending/webhooks/synchronous-events/overview.mdx)
+## Tech Agnostic
 
-- **Dashboard extensions.** [Extend admin](/developer/extending/apps/overview.mdx) functionality using any front-end stack embeddable into _iframes_.
+While Saleor's engine is written in Python, it makes almost no assumptions about your tech stack. Saleor's API and the extension mechanism utilize GraphQL transported over either HTTP or message queues. Hence, the only requirement we impose is that your code be able to make or receive HTTP requests.
```
peelar, line 11:
```suggestion
While Saleor's engine is written in Python, it makes almost no assumptions about your tech stack. Saleor's API and the extension mechanism utilize GraphQL transported over either HTTP or message queues. Hence, the only requirement we impose is that your code is able to make or receive HTTP requests.
```

docs/overview/why-saleor/extensibility.mdx
```diff
@@ -1,48 +1,81 @@
...
 
-- **Granular permission control.** [Granular permissions](/developer/permissions.mdx) for access tokens.
+Whether you work with Python, TypeScript, PHP, Java, C#, Elixir, or Rust, you can integrate with Saleor. And chances are there is already a GraphQL client library you can use with enough tooling to provide a fantastic developer experience.
 
-- **GraphQL oriented.** Provide GraphQL queries to shape the data provided via Webhooks.
+At Saleor, we always strive to build on existing standards. This way, you get the best interoperability and the benefit of existing tools and experience, and no one has to train specifically to become a "Saleor developer."
```
peelar, line 15:
```suggestion
At Saleor, we always strive to build on existing standards. This way, you get the best interoperability and the benefit of existing tools and experience, and no one has to train specifically to become a "Saleor developer".
```

docs/overview/why-saleor/extensibility.mdx
```diff
@@ -1,48 +1,81 @@
...
-### Why not plugin architecture?
+## Modeling
 
-The plugin architecture is popular among open-source projects; classic examples would be Magento or Shopify.
-Such a model has direct coupling with the platform,
-which can be convenient to get started quickly but can also lead to many issues.
+Extensibility starts with mapping your business needs to data shapes. While there are stores where the product is nothing but a name, a picture, and a two-sentence description, we recognize that it's rarely the case.
 
-**Benefits of decoupled architecture:**
+Saleor's product type system allows you to create named schemas for different product types, each having a different set of fields of various types. This guarantees that your books always have a page count that is a number and that your drinks have a list of ingredients, each linked to its own entity.
```
peelar, line 21:
```suggestion
Saleor's product type system allows you to create named schemas for different product types, each having a different set of attributes of various types. This guarantees that your books always have a page count that is a number and that your drinks are linked with their ingredients.
```

**style (non-blocking):**
The rationale behind the first proposed change is to use the vocabulary that may be already familiar to Saleor users.

With the second one, I am trying to propose a bit simpler structure. To me, it currently reads like there are three things: drinks, ingredients and some entity. But maybe we want to emphasize that ingredient is a separate entity.

docs/overview/why-saleor/extensibility.mdx
```diff
@@ -1,48 +1,81 @@
...
 
-- **Technology agnostic.**
-  Technology stack freedom allows you to pick the best tool for the job and access a larger talent pool.
+Entities, like your drink ingredients, also have named shapes, making sure the allergen field is a boolean flag and that the rich text description is present.
```
peelar, line 23: **thought:**

Similarly as above, I suggest we try to stick to Saleor vocabulary because I find words such as "entities" and "named shapes" a bit confusing.

docs/overview/why-saleor/extensibility.mdx
```diff
@@ -1,48 +1,81 @@
...
 
-- **Improved security.** Tight coupling increases security risks whereas decoupled architectures reduce the attack surface.
+Our "Commerce as Code" approach gives you tools to programmatically control these attributes and schemas, so the shape of your data can be version-controlled along with the source code of the storefronts and integrations that rely on the fields.
 
-- **Independent deployment.** You can deploy your microservices independently.
+For programmatic use, almost all objects in Saleor support metadata, which allows for the storage and retrieval of practically any key-value data.
```
peelar, line 27: **suggestion (non-blocking):**

I will once again challenge the idea of not referencing at least some of the existing, related documents. Here, we could link to the metadata docs, like so:

```suggestion
For programmatic use, almost all objects in Saleor support [metadata](/api-usage/metadata.mdx), which allows for the storage and retrieval of practically any key-value data.
```

It would look strange as the only link in the document, so we could think about adding a couple more.

peelar (changes requested):

peelar (dismissed):

docs/overview/why-saleor/extensibility.mdx
patrys: I agree with the "attributes" change, but the second change is semantically different. My version invokes a many-to-many relationship through an orderable list of references; yours implies a many-to-one relationship.

docs/overview/why-saleor/extensibility.mdx
patrys: Saleor vocabulary is "Page" and "Menu" and is about to be replaced because it's even more confusing.

peelar (approved):

peelar (approved):

PR #1563, opened 2025-05-05
Add robots.txt file

krzysztofwolski:
## Description

Add missing `robots.txt`. All crawlers welcome :)

## Checklist

- [*] I have read and followed the [guidelines](../GUIDELINES.md)

### DIFF
--- a/docs/cloud/overview.mdx
+++ b/docs/cloud/overview.mdx
@@ -7,7 +7,7 @@ sidebar_title: Overview
 
 Saleor cloud provides free test projects for experimentation and development purposes.
 
-Signup at [cloud.saleor.io](https://cloud.saleor.io/) to get started for free.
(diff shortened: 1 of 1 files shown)


NyanKiyoshi:
@krzysztofwolski are you able to make it so preview deployments aren't indexed? (unless they are already excluded? - I couldn't find an exclusion on my end)

krzysztofwolski:
@NyanKiyoshi Vercel is kind enough to cover this by default: https://vercel.com/guides/are-vercel-preview-deployment-indexed-by-search-engines
X-Robots-Tag is set to 'noindex' for preview deployments making them ignored by the crawlers.
So we are good here :)

static/robots.txt
```diff
@@ -0,0 +1,3 @@
+User-agent: *
+Disallow:
```
krzysztofzuraw, line 2: Don't we want:
```
User-agent: *
Allow: /
```
I mean it is more clear that we allow all in such syntax 🤔

krzysztofwolski: Sure. Both rules would work, but your right about clarity. Update pushed :)

PR #1567, opened 2025-05-06
Update prices page with max price value

IKarbowiak:
## Description

Add info about max price value on prices page.
Related PR: https://github.com/saleor/saleor/pull/17678

## Checklist

### DIFF
--- a/docs/api-usage/prices.mdx
+++ b/docs/api-usage/prices.mdx
@@ -4,6 +4,7 @@ title: Prices
 
 Saleor API has four different types that represent prices. The most basic type consists of a currency code and a numeric amount. The most complex one can define a range of prices, including net, gross, and tax amounts.
 The numeric amount precision is determined by the precision of the price currency. For example, USD has a precision value of 2, such as $1.00.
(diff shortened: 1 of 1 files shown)


docs/api-usage/prices.mdx
```diff
@@ -4,6 +4,7 @@ title: Prices
 
 Saleor API has four different types that represent prices. The most basic type consists of a currency code and a numeric amount. The most complex one can define a range of prices, including net, gross, and tax amounts.
 The numeric amount precision is determined by the precision of the price currency. For example, USD has a precision value of 2, such as $1.00.
+Regardless of the currency, the maximum price value that can be provided is 99,999,999,999. If this value is exceeded, an error will be raised.
```
lkostrowski, line 7: `,` is confusing here, what about decimal place? How I understand it will be max digits, including decimal places?

krzysztofzuraw: I think it depends if you use `,` and decimal places separator or `.` Docs: https://docs.oracle.com/cd/E19455-01/806-0169/overview-9/index.html
I think we can use `99 999 999 999 999`

krzysztofzuraw (approved):

PR #1578, opened 2025-05-15
Add page with app error explanations

witoszekdev:
This PR adds new page for app installation errors. It provides detailed explanation why error occured, and how to debug this.

This idea came from adding links to dashboard, when errors occur - instead of linking to a generic page which describes, for example app's manifest user will have more smooth experience by directly addressing their issue.

It will be used in Saleor Dashboard to link here, but without listing it as a navigation item (think: React docs, that link from specific error in browser's console). This way user can learn more about error and how to fix it.


docs/developer/extending/apps/developing-apps/app-error-codes.mdx
```diff
@@ -0,0 +1,225 @@
...
+
+Below you'll find explanations for `AppErrorCode` values that can be returned during various app-related operations in Saleor, such as installing an app, fetching its manifest, updating an app, or deleting it. Each section explains what the error means, when it can happen, and what you should check or do to resolve it.
+
+## `AppErrorCode.INVALID_URL_FORMAT`
+This error is raised when a URL is malformed or uses an unsupported scheme.
+It can occur for field `manifestUrl` (when fetching app's manifest) or for any URL field within the manifest content
+(such as `tokenTargetUrl`, extension URLs, webhook `targetUrl`, or brand logo URLs).
+
+- 🧐 **Cause**: A URL (either `manifestUrl` or a URL inside the manifest) is invalid, malformed, or uses an unsupported scheme for its context. For example:
+  - `manifestUrl` supports only `http://` or `https://`.
```
korycins, line 13: Are you sure that we always support  `http` / `https`? I think that this is controller by the environment 🤔  In debug mode (local env) `http` works as expected, but drops the `https` manifests. When hosted, we will drop `http` request but accept `https`.
Maybe worth to check this/mention?

docs/developer/extending/apps/developing-apps/app-error-codes.mdx
```diff
@@ -0,0 +1,225 @@
...
+  - Inspect the full manifest content.
+  - Check error details, which should indicate the specific required field(s) that are missing.
+  - Refer to Saleor's [App Manifest documentation](/developer/extending/apps/architecture/manifest.mdx) to ensure all required fields are present.
+
+## `AppErrorCode.UNIQUE`
+This error means the `id` field (identifier) in the manifest is already used by another app in your Saleor instance.
+
+- 🧐 **Cause**: The app identifier specified in the manifest is not unique within the Saleor instance.
+- 💡 **Debugging Tips**:
+  - Check if an app with the same `id` is already installed.
```
korycins, line 101: I think that we also take into account the ongoing installation - maybe worth to mention it here

docs/developer/extending/apps/developing-apps/app-error-codes.mdx
witoszekdev: I've checked it on production with URL: `http://localhost:3000` and we return error code `AppErrorCode.INVALID`, not this one, since Saleor couldn't fetch the host - validation didn't pick this up.

It makes sense because we use:
- `class AppURLValidator(URLValidator)` - which uses Django's URLValidator, which [by default supports `http` protocol](https://docs.djangoproject.com/en/5.2/ref/validators/#django.core.validators.URLValidator.schemes)
- for `targetUrl` we override this to allow awssqs and gcppubsub protocols:

```
    target_url_validator = AppURLValidator(
        schemes=["http", "https", "awssqs", "gcpubsub"]
    )
```

witoszekdev: @korycins maybe we should change this validation, as you mentioned since it doesn't make sense to connect to local ports on production

docs/developer/extending/apps/developing-apps/app-error-codes.mdx
witoszekdev: yup you were right, updated that part

andrzejewsky (approved):

PR #1587, opened 2025-05-23
fix: missing sidebar in AI docs

NyanKiyoshi:
This fixes the missing sidebar in our AI documentation.

| Before | After |
| --- | --- |
| Screenshot 2025-05-23 at 11 58 26 AM | Screenshot 2025-05-23 at 11 58 15 AM |

### DIFF
--- a/docs/developer/ai.mdx
+++ b/docs/developer/ai.mdx
@@ -1,5 +1,6 @@
 ---
 tags: [AI, "Artificial Intelligence"]
+displayed_sidebar: main
 ---
 
 # Using AI with Saleor


patrys (changes requested):
It would be easier to change the "ref" (which means an external, non-binding link) in the main sidebar into a regular "doc" with a custom icon. It would be automatically assigned to the sidebar.

patrys:
@NyanKiyoshi `ref` is a shortcut for creating a `type: "ref"` entry, it does not make sense to have it return any other type. Just replace `ref("developer/ai", "AI", "star")` with `{type: "doc", id: "developer/ai", customProps: {icon: "star"}}` (you don't need to pass a label, it should be extracted from the document).

NyanKiyoshi:
> @NyanKiyoshi `ref` is a shortcut for creating a `type: "ref"` entry, it does not make sense to have it return any other type. Just replace `ref("developer/ai", "AI", "star")` with `{type: "doc", id: "developer/ai", customProps: {icon: "star"}}` (you don't need to pass a label, it should be extracted from the document).

@patrys Thanks! Fixed at 9d875db05decafd1e6d843aa9a6a3c75c639699d

patrys (approved):

krzysztofzuraw (approved):

PR #1623, opened 2025-06-23
Docusaurus 3.8

krzysztofwolski:
## Description

- bump docusaurus version. Theoretically it should improve builds, but vercel seems to be ignoring the cache :( On local env improvement is clearly visible tho (helps a lot when debugging locally - 20s builds are so nice :) )
- as usual, took a look at deps and removed/updated what I could
- sidebar component updated to the version from official repo
- adjusted top margin which looked bad on mobile
- Change breakpoint for chapters component - previous was causing text overlap

## Checklist

- [x] I have read and followed the [guidelines](../GUIDELINES.md)

### DIFF
--- a/components/Chapters/styles.module.css
+++ b/components/Chapters/styles.module.css
@@ -14,7 +14,7 @@
   }
 }
 
-@container (width < 200px) {
+@container (width < 400px) {
   .grid {
     grid-template-columns: repeat(1, minmax(100px, 200px));
   }


package.json
```diff
@@ -29,18 +29,18 @@
...
-    "@docusaurus/preset-classic": "3.7.0",
-    "@docusaurus/theme-classic": "3.7.0",
-    "@docusaurus/theme-mermaid": "3.7.0",
-    "@docusaurus/types": "3.7.0",
+    "@docusaurus/core": "^3.8.1",
+    "@docusaurus/faster": "^3.8.1",
+    "@docusaurus/preset-classic": "^3.8.1",
+    "@docusaurus/theme-classic": "^3.8.1",
+    "@docusaurus/theme-mermaid": "^3.8.1",
+    "@docusaurus/types": "^3.8.1",
```
witoszekdev: issue: We should pin version numbers otherwise it will be automatically bumped to next minor. We might also use `~` for auto patch updates

```suggestion
    "@docusaurus/core": "3.8.1",
    "@docusaurus/faster": "3.8.1",
    "@docusaurus/preset-classic": "3.8.1",
    "@docusaurus/theme-classic": "3.8.1",
    "@docusaurus/theme-mermaid": "3.8.1",
    "@docusaurus/types": "3.8.1",
```

witoszekdev (changes requested):

package.json
krzysztofwolski: Their automatic script seems to not be respecting version pinning :< Thanks for the catch. Change pushed 🙌

aniav (approved):

src/theme/DocRoot/Layout/Sidebar/index.js
```diff
@@ -43,57 +50,39 @@ export default function DocRootLayoutSidebar({
...
           return;
         }
+
         if (hiddenSidebarContainer) {
           setHiddenSidebar(true);
         }
       }}
     >
-      <SwitchTransition>
-        <CSSTransition
```
lkostrowski, line 69: why removed animations?

components/AskAI/index.jsx
lkostrowski, line 1: what was it and why removed?

src/theme/DocRoot/Layout/Sidebar/index.js
```diff
@@ -1,3 +1,10 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
```
kzuraw, line 2: question: Is this copied from Facebook?

package.json
```diff
@@ -49,20 +49,29 @@
...
+    "react-dom": "^19.0.0"
+  },
+  "browserslist": {
+    "production": [
+      ">0.5%",
+      "not dead",
+      "not op_mini all"
+    ],
+    "development": [
+      "last 3 chrome version",
```
kzuraw: nitpick: maybe we can have 2 last versions of all browsers - as this means polyfills for Safari 13 (from 2020)

src/theme/DocRoot/Layout/Sidebar/index.js
krzysztofwolski: Yes, from their main template repo.

src/theme/DocRoot/Layout/Sidebar/index.js
krzysztofwolski: There were changes in the upstream implementation of the theme and copied it.
Found no animation much less distracting and snappier.

components/AskAI/index.jsx
krzysztofwolski: It was oooold integration with service providing chat-like conversations. At some point we turned it off because it was producing non satisfactory results. Checked it lately, and they seems to no longer offer such free service so decided its time to clean it up.

package.json
krzysztofwolski: Updated

components/AskAI/index.jsx
krzysztofwolski: And for the context - animation on mobile for showing/hiding menu stays. Animation discussed here was triggering when going deeper into navigation tree or going back.

witoszekdev (approved):

PR #1631, opened 2025-07-08
Add info about increased query cost

korycins:
## Checklist

### DIFF
--- a/docs/upgrade-guides/core/3-21-to-3-22.mdx
+++ b/docs/upgrade-guides/core/3-21-to-3-22.mdx
@@ -1,0 +1,64 @@
+---
+title: Upgrading From 3.21 To 3.22
+sidebar_position: 0
+---
+
+:::info
+To follow the zero-downtime strategy when upgrading to 3.21, **It is recommended to first migrate to latest 3.20.X** and turn on the Celery worker to process all data migrations asynchronously.
+Otherwise, you will need to downtime your solution to ensure correct data migration.
+:::
+
+## Increased query costs for specific fields
+
+To further enhance API performance and system stability, the query cost for several fields has been increased:
+
+- `Page.attributes`: **from 1 to 100**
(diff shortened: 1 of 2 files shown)


kadewu (dismissed):

docs/upgrade-guides/core/3-21-to-3-22.mdx
```diff
@@ -0,0 +1,64 @@
...
+    }
+  }
+}
+```
+
+### Example: Query cost calculation
+
+The following query used to return a cost of 200, but with the updated cost model, it now results in a cost of 10 100.
+
+10 100 = 100 (pages.first) + 100 (pages) × 100 (attributes)
```
NyanKiyoshi: English uses a comma as the delimiter:

```suggestion
The following query used to return a cost of 200, but with the updated cost model, it now results in a cost of 10,100.

Where 10,100 is 100 (pages.first) + 100 (pages) × 100 (attributes)
```

(caused it to be initially confusing to read until I realized it was supposed to represent the number 10100)

I'm not sure about keeping the comma though as it's a small number, maybe just putting `10100` would be enough? (should be clearer for all countries in the world, however it's still slightly "weird looking" due to looking like a binary)

IKarbowiak (dismissed):

docs/upgrade-guides/core/3-21-to-3-22.mdx
```diff
@@ -0,0 +1,64 @@
...
+  "extensions": {
+    "cost": {
+      "requestedQueryCost": 100,
+      "maximumAvailable": 50000
+    }
+  }
+}
+```
+
+### Example: Query cost calculation
```
lkostrowski, line 43: shouldn't we put this to main docs as well?

korycins: added

PR #1657, opened 2025-08-11
Refund reasons references

lkostrowski:
Upcoming in 3.22

### DIFF
--- a/docs/developer/attributes/api.mdx
+++ b/docs/developer/attributes/api.mdx
@@ -45,6 +45,16 @@ To fetch the list of attributes, use the `attributes` query:
         slug
         inputType
         entityType
+        referenceTypes {
+          ...on ProductType {
+            id
+            slug
+          }
+          ...on PageType {
+            id
+            slug
+          }
+        }
         unit
         choices(first: 5) {
           edges {
@@ -65,7 +75,9 @@ The `attributes` field supports pagination and filtering, e.g., searching attrib
 - `name` - Name of the attribute, use it to render attribute in the frontend.
 - `slug` - A unique attribute handle you can use to fetch an attribute or reference it in filters.
 - `inputType` - Represents the input type.
-- `entityType` - Used only with `REFERENCE` input types; represents the type of entities associated as references, e.g., `PAGE` or `PRODUCT`.
+- `entityType` - Used only with `REFERENCE` and `SINGLE_REFERENCE` input types; represents the type of entities associated as references, e.g., `PAGE` or `PRODUCT`.
+- `referenceTypes` - Allowed target types for reference attributes. For attributes with the `PRODUCT` or `PRODUCT_VARIANT` entity type, this lists allowed product types;
+for attributes with the `PAGE` entity type, this lists allowed page types. Not supported for `CATEGORY` or `COLLECTION` reference attributes.
 - `unit` - Used only with `NUMERIC` input types; represents the optional measurement unit.
 - `choices` - Used only with `DROPDOWN`, `MULTISELECT` and  `SWATCH`; a list of predefined choices you can use as values. The `choices` field supports pagination and filtering, e.g., searching choices by name.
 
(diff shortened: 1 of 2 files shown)


docs/developer/payments/refunds.mdx
```diff
@@ -291,3 +303,39 @@ Below is an example order with multiple transactions and `totalRemainingGrant` a
...
+
+
+
+## Refund settings
+
+:::note
+Available from 3.22
+:::
+
+While `reason` field is optional and remains a plain text field, the `reasonReference` can be configured:
```
IKarbowiak: IMO we should firstly add short explanation what is this refund setting, and why it's usefull.

docs/developer/payments/refunds.mdx
```diff
@@ -291,3 +303,39 @@ Below is an example order with multiple transactions and `totalRemainingGrant` a
...
+
+```graphql
+mutation UpdateRefundSettings($id: ID) {
+    refundSettingsUpdate(input: {
+      reasonReferenceType: $id
+    }) {
+        reasonReferenceType
+    }
+}
+```
```
IKarbowiak: You should also attach the example variables and show the response of this mutations. Use `Tab` for it. You can see the examples here: https://github.com/saleor/saleor-docs/pull/1674

docs/developer/payments/refunds.mdx
```diff
@@ -291,3 +303,39 @@ Below is an example order with multiple transactions and `totalRemainingGrant` a
...
+
+- input `reasonReferenceType` - the ID of a Model Type. All models that are instances of this type will be available to reference in the reasons.
```
IKarbowiak: `the ID of a Model Type.`
- `Model Type` is `Page Type` in the `API`, so probably we should mentioned both until it changed, especially that you describe the mutation

docs/developer/payments/refunds.mdx
```diff
@@ -42,7 +42,9 @@ mutation TransactionRequestAction($id: ID, $actionType: TransactionActionEnum!,
 {
   "id": "VHJhbnNhY3Rpb25JdGVtOjljY2NkYTYyLTllMjktNGE0OC05NzIyLWRlYzAwOTRmZmY5Yg==",
   "actionType": "REFUND",
-  "amount": "10"
+  "amount": "10",
+  "refundReason": "Customer requested a refund"
```
korycins:
```suggestion
  "refundReason": "Customer requested a refund",
```

docs/developer/payments/refunds.mdx
```diff
@@ -334,3 +343,70 @@ Below is an example order with multiple transactions and `totalRemainingGrant` a
   It also means that `totalChargedAmount` includes `amountChargePending`, and `totalRefundedAmount` includes `amountRefundPending`.
 - Authorized amounts, including `amountAuthorized` and `amountAuthorizePending`, are also considered processed. These amounts will contribute to the overall processed totals and impact the `overchargedAmount`.
   :::
+
+
+## Refunds configuration
```
korycins: Maybe worth to change the title  of this section?

My first impression was: “We have Refund configuration + Available from 3.22 -> does that mean I couldn’t process refunds before?”

It might be worth making the section title more explicit. For example:
- Change “Refunds configuration” -> “Refund reasons configuration”, or
- Keep “Refunds configuration”, but move the “Available from 3.22” note specifically under the refund reason configuration part.

docs/developer/payments/refunds.mdx
```diff
@@ -334,3 +343,70 @@ Below is an example order with multiple transactions and `totalRemainingGrant` a
...
+      "pageTypeId": "your-id"
+    }
+    
+    ```
+    </TabItem>
+</Tabs>
+
+Now, every time a *staff user* tries to create a refund via `transactionActionRequest` or `orderGrantRefundCreate`/`orderGrantRefundUpdate` mutations (or via Dashboard), they will be required to provide a reason reference. The existing "reason" field will still be available, but it will be optional.
+
+Note that permitted `App` will be able to omit this setting. For the User, the ValidationError will be raised if the reason reference is not provided. App can omit this field for more powerful workflows.
```
korycins:
```suggestion
Note that permitted `App` will be able to omit this setting. For the *staff user*, the ValidationError will be raised if the reason reference is not provided. App can omit this field for more powerful workflows.
```
(to align with the format used in 390)

korycins: OR
```suggestion
Note that permitted `App` will be able to omit this setting. For the `User`, the ValidationError will be raised if the reason reference is not provided. App can omit this field for more powerful workflows.
```
(to treat User word as `User` type - align with `App` usage)

korycins: I would consider placing it inside:
```
:::note
<>
:::
```
To keep this information more visible (as it says that the validation works differently for the App)

docs/developer/payments/refunds.mdx
```diff
@@ -334,3 +343,70 @@ Below is an example order with multiple transactions and `totalRemainingGrant` a
...
+  refundReasonReferenceClear {
+    errors {
+      message
+    }
+  }
+}
+```
+
+:::note
+You may have seen we start referencing `Page` as a `Model`. While in the API we still use `Page` name for backwards compatibility, we will be renaming it to `Model` in the future.
```
korycins: If you want to include this information, I think it should appear before the first mention of the Model type. Currently I might spend time trying to figure out what Model refers to in the sections above.

docs/developer/payments/refunds.mdx
```diff
@@ -42,7 +42,9 @@ mutation TransactionRequestAction($id: ID, $actionType: TransactionActionEnum!,
 {
   "id": "VHJhbnNhY3Rpb25JdGVtOjljY2NkYTYyLTllMjktNGE0OC05NzIyLWRlYzAwOTRmZmY5Yg==",
   "actionType": "REFUND",
-  "amount": "10"
+  "amount": "10",
+  "refundReason": "Customer requested a refund"
+  "refundReasonReference": "xXyzaXdaxbX" # ID of a model representing the reason for the refund
```
IKarbowiak: It's better to use actual base64 id, that decoded to `Page`, like:
```suggestion
  "refundReasonReference": "UGFnZTox"  # ID of a model representing the reason for the refund
```
image

docs/developer/payments/refunds.mdx
```diff
@@ -138,7 +143,8 @@ mutation OrderGrantRefundCreate($id: ID!, $input: OrderGrantRefundCreateInput!)
     ],
     "grantRefundForShipping": true,
     "amount": 10,
-    "reason": "Returned by customer"
+    "reason": "Returned by customer",
+    "reasonReference": "xXyzaXdaxbX" # ID of a Page
```
IKarbowiak: Same here:
```suggestion
    "reasonReference": "UGFnZTox" # ID of a Page
```

Also here, you are saying `ID of a Page` and in the `TransactionRequestAction` part is `ID of a Model` I think it should be consistent. Like `ID of a Model (Page)`

docs/developer/payments/refunds.mdx
```diff
@@ -334,3 +343,70 @@ Below is an example order with multiple transactions and `totalRemainingGrant` a
...
+
+:::note
+Available from 3.22
+:::
+
+Staff users with `MANAGE_SETTINGS` permission can access additional refunds configuration available in Dashboard and via API.
+
+Configuration is global and reflected on every channel.
+
+Currently, `RefundSettings` type controls a single behavior: whether the "reason reference" is required when the refund is created or updated.
```
IKarbowiak: Maybe use the variable that is available in inputs
```suggestion
Currently, `RefundSettings` type controls a single behavior: whether the `reasonReference` is required when the refund is created or updated.
```

docs/developer/payments/refunds.mdx
```diff
@@ -334,3 +343,70 @@ Below is an example order with multiple transactions and `totalRemainingGrant` a
...
+
+### Setting refund reason type requirement
```
IKarbowiak: We are using Chicago style convention, so each word should be capitalised:
```suggestion
### Setting Refund Reason Type Requirement
```

PR #1774, opened 2026-04-09
Update API reference to 3.23

lkostrowski:
Large negative diff is related to graphql-markdown package, which in the update (since last regenerate) reduced dead code (react component added to every markdown file)


peelar (approved):

patrys:
@lkostrowski This broke the `<Details>` component, here's an example of what it looks like now: https://docs.saleor.io/api-reference/authentication/mutations/external-logout

PR #1786, opened 2026-05-12
Payment status clarification

szczecha:
## Description

The topic comes back like a boomerang, so I decided to add a warning in the docs that this field is not accurate when using transactions.
We can rearrange this section in a separate PR.
https://saleor-docs-cke1oydcj-saleorcommerce.vercel.app/developer/payments/lifecycle

## Checklist

- [x] I have read and followed the [guidelines](../GUIDELINES.md)

### DIFF
--- a/docs/developer/payments/lifecycle.mdx
+++ b/docs/developer/payments/lifecycle.mdx
@@ -3,7 +3,7 @@ title: Lifecycle
 sidebar_label: Lifecycle
 ---
 
-## Payment status on Order
+## Payment Status on Order
 
 :::caution
 `paymentStatus` is a legacy field inherited from the old Payments API (payment plugins) 
@@ -68,7 +68,7 @@ Creating the refund will change the status to:
 `Checkout` doesn't have a `paymentStatus` field.
 :::
 
-## Authorize status
+## Authorize Status
 
 ```mermaid
 stateDiagram-v2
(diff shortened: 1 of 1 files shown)


docs/developer/payments/lifecycle.mdx
```diff
@@ -3,7 +3,15 @@ title: Lifecycle
...
-## Payment status on Order 
+## Payment Status on Order
+
+:::caution
+`paymentStatus` is a legacy field inherited from the old Payments API (payment plugins) 
+It attempts to map `TransactionItem` state into a single combined field. 
+It works correctly in most cases, but certain edge cases fall outside what a single field can accurately represent.
+
+For a precise view of payment state, use [`authorizeStatus`](#authorize-status) and [`chargeStatus`](#charge-status) instead — they are calculated directly from `TransactionItem` amounts.
+:::
```
NyanKiyoshi, line 14: Can we deprecate that field and update the description so that it's clear to users? (i.e., they don't need to check the docs learn about it)


That's what I see in `main`:

Image

docs/developer/payments/lifecycle.mdx
```diff
@@ -36,10 +44,10 @@ stateDiagram-v2
  PARTIALLY_REFUNDED --> FULLY_REFUNDED
 ```
 
-[`Order.paymentStatus`](/api-reference/orders/objects/order.mdx#payment-status) represents the combined payment and refund status of the order. It's calculated based on:
+[`Order.paymentStatus`](/api-reference/orders/objects/order.mdx#payment-status) attempts to consolidate both payment and refund state into a single field. It is calculated based on:
```
NyanKiyoshi, line 47: "attempts to" - can we describe in which case it doesn't work?

stmpn (approved):

PR #1811, opened 2026-07-21
Fix auto-update api reference

lkostrowski:
# Problem

1. No auto-run, our reference is stale, needs to be updated manually
2. Existing workflow is broken

# Solution

1. Fix workflow 
2. Make it run daily
3. Cache - check if there is new release and run only then

# Proof
https://github.com/saleor/saleor-docs/actions/runs/29821293110


.github/workflows/update-api-reference.yml
```diff
@@ -2,16 +2,61 @@ name: Update 3.x docs to the latest Saleor release
...
+        with:
+          path: .last-release
+          key: saleor-release-${{ github.run_id }}
+          restore-keys: saleor-release-
+
+      - name: Check for new release
+        id: check
+        run: |
+          last=$(cat .last-release 2>/dev/null || echo "")
+          if [[ "$last" == "${{ steps.release.outputs.tag }}" ]]; then
```
NyanKiyoshi: `${{ steps.release.outputs.tag }}` contains arbitrary user input, you must move that to an env var

.github/workflows/update-api-reference.yml
```diff
@@ -2,19 +2,70 @@ name: Update 3.x docs to the latest Saleor release
...
+        id: release
+        run: |
+          tag=$(curl -s https://api.github.com/repos/saleor/saleor/releases/latest | jq -r '.tag_name')
+          echo "tag=$tag" >> "$GITHUB_OUTPUT"
+          echo "Latest Saleor release: $tag"
+
+      # Rolling cache: restore-keys picks up the most recent previous cache,
+      # and we save a fresh one each run keyed by run_id (cache keys are immutable).
+      - name: Restore last scanned release
+        uses: actions/cache/restore@0057852bfaa89a56745cba8c7296529d2fc39830 # v4.3.0
```
NyanKiyoshi: 4.3.0 is ancient:

```suggestion
        uses: actions/cache/restore@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
```

.github/workflows/update-api-reference.yml
```diff
@@ -50,16 +88,23 @@ jobs:
...
-            Pull request auto-generated by bot
+            Pull request auto-generated by bot for Saleor release ${{ env.TAG }}.
           labels: |
             :robot: bot
+          team-reviewers: engineering
+
+      # Record the scanned tag only after a successful run so a failure retries next night.
+      - name: Save last scanned release
+        run: echo "$TAG" > .last-release
+      - uses: actions/cache/save@0057852bfaa89a56745cba8c7296529d2fc39830 # v4.3.0
```
NyanKiyoshi:
```suggestion
      - uses: actions/cache/save@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
```

lkostrowski:
Fixed, also changed github bot to saleor bot, so workflows can run automatically on the output PR

.github/workflows/update-api-reference.yml
```diff
@@ -88,9 +89,22 @@ jobs:
...
+      # and the push is allowed to touch workflow files.
+      - name: Generate GitHub access token
+        id: generate-token
+        env:
+          VAULT_URL: ${{ secrets.VAULT_URL }}
+          VAULT_JWT: ${{ secrets.VAULT_JWT }}
+        run: |
+          token=$(curl -s --request GET --url "$VAULT_URL" --header "Authorization: JWT $VAULT_JWT" | jq -r .token)
+          echo "::add-mask::$token"
+          echo "token=$token" >> "$GITHUB_OUTPUT"
```
NyanKiyoshi: 3 issues:
1. This should be using the `saleor-internal-actions`
2. The secret should be inside a GitHub environment so it cannot be accessible to other workflows
3. This secret should be only used inside a separate job that doesn't contain any NPM dependencies

NyanKiyoshi (changes requested):

.github/workflows/update-api-reference.yml
lkostrowski: done

.github/workflows/update-api-reference.yml
```diff
@@ -50,16 +85,64 @@ jobs:
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+        with:
+          ref: ${{ github.event.repository.default_branch }}
+          persist-credentials: false
+
+      # Replace the whole dir so docs removed in the new schema are dropped too.
+      - name: Apply generated API reference
+        run: rm -rf docs/api-reference
```
NyanKiyoshi, line 113: Unnecessary `--force`:

```suggestion
        run: rm -r docs/api-reference
```

NyanKiyoshi (approved):

PR #1815, opened 2026-08-03
Add docs for Customer Types feature

wcislo-saleor:
> [!CAUTION]
> ✅ Rebase after https://github.com/saleor/saleor-docs/pull/1818 is merged.
> ✅ Update API reference with version `3.23.26` then.


docs/developer/users/customer-types.mdx
```diff
@@ -0,0 +1,205 @@
+---
+title: Customer Types and Attributes
+description: Group customers with customer types and attach structured attribute data to them.
+---
+
+**Added in Saleor 3.23.**
+
+Customer types let you segment customers and attach structured data to their accounts. A customer type is a named collection of attributes — the same [attribute system](/developer/attributes/overview.mdx) used by product and page types — so you can model concepts like loyalty tiers, company details for B2B accounts, or onboarding metadata, and have them validated and queryable through the GraphQL API.
+
+Every customer belongs to exactly one customer type. The type determines which attributes can be set on the customer and which attribute values are visible when reading the customer.
```
przlada:
```suggestion
Every customer belongs to exactly one customer type. The type determines which attributes can be set on the customer and which attribute values are returned when querying the customer.
```

docs/developer/users/customer-types.mdx
```diff
@@ -0,0 +1,205 @@
...
+    assignedAttribute(slug: "loyalty-level") {
+      attribute {
+        slug
+      }
+    }
+  }
+}
+```
+
+`assignedAttributes` returns the attributes of the user's current customer type together with their values. Staff with `MANAGE_USERS` see all of them, while the account owner sees only attributes visible in the storefront.
```
przlada:
```suggestion
`assignedAttributes` returns the attributes of the user's current customer type together with their values. Staff or apps with `MANAGE_USERS` see all of them, while the account owner sees only attributes visible in the storefront.
```

docs/developer/users/customer-types.mdx
```diff
@@ -0,0 +1,205 @@
...
+    }
+  }
+}
+```
+
+Filtering by the default customer type also matches users that were never explicitly assigned a type.
+
+## Attribute Value Lifecycle
+
+Attribute values are stored independently of the customer type they were written under. Visibility is evaluated at read time against the user's current type:
```
przlada:
```suggestion
Attribute values are stored independently of the customer type they were written under. Visibility is evaluated at read and filter time against the user's current type:
```

docs/developer/users/customer-types.mdx
przlada, line 1: I think that there is one more behavior worth documenting: customer attribute values are excluded from the translations API - `translations(kind: ATTRIBUTE_VALUE)` skips them. Since product and page attribute values are translatable, it can be assumed that customer values are too. Maybe a short note in this section would do, something like:                                                                            

> Values of customer attributes may contain personal data and are excluded from the translations API - they are not listed as translatable content and cannot be translated

wcislo-saleor: Added.

przlada (approved):

PR #1820, opened 2026-08-12
fix(style): format .mdx files

NyanKiyoshi:
This uses & configures prettier in order to automatically format MDX files to have a consistent style & help with readability.

## Checklist

- [x] I have read and followed the [guidelines](../GUIDELINES.md)

PR #1823, opened 2026-08-14
fix(style): format .mdx files under docs/{cloud,demos,overview}/

NyanKiyoshi:
Partial apply of #1820 whiches uses & configures prettier in order to automatically format MDX files to have a consistent style & help with readability. This changeset solely focuses on `./docs/{cloud,demos,overview}/`, then follow-up changesets will be opened for other directories.

## Checklist


.prettierignore
```diff
@@ -3,3 +3,6 @@ build/
 .docusaurus/
 
 !docs/api-usage/**/*.mdx
+!docs/cloud/**/*.mdx
+!docs/demos/**/*.mdx
+!docs/overview/**/*.mdx
```
lkostrowski, line 8: is it temporary? If so, maybe comment?

lkostrowski (approved):

.prettierignore
NyanKiyoshi: Yes, it's temporary until we are done reformatting everything. I can add a comment in a follow-up PR

